PR #25, opened 2020-04-28
Connect to postgres

ibeckermayer:
The db query endpoint should work now. Updated the scripts a bit, see the README for running locally.


permission-portal-server/config.js
```diff
@@ -5,9 +5,15 @@ dotenv.config();
...
-const connectionString = `postgresql://${process.env.DB_USER}:${process.env.DB_PASSWORD}@${process.env.DB_HOST}:${process.env.DB_PORT}/${process.env.DB_DATABASE}`
+const connectionString = `postgres://${process.env.DB_USER}:${process.env.DB_PASSWORD}@${process.env.DB_HOST}:${process.env.DB_PORT}/${process.env.DB_DATABASE}`;
 
 export const pool = new Pool({
-  connectionString: isProduction ? process.env.DATABASE_URL : connectionString,
-  ssl: isProduction
-});
\ No newline at end of file
+  connectionString: isProduction
+    ? process.env.HEROKU_POSTGRESQL_BLUE_URL
```
freyley, line 12: Heroku can provide multiple different POSTGRESQL_<COLOR> urls, but the main one will be DATABASE_URL. We should probably use that instead of this.

permission-portal-server/routes/organization.js
```diff
@@ -11,15 +12,16 @@ organizationRouter.use(bodyParser.json());
 
 const getOrgs = async (req, res, next) => {
   try {
-    const results = await pool.query("SELECT * FROM Organizations");
+    const client = await pool.connect();
+    const results = await client.query('SELECT * FROM "Organization"');
```
freyley, line 16: Do folks have strong opinions on ORMs? I'm generally a fan of ones that have enough power to let you get under the hood when you need to. I'm not familiar enough with Sequelize - https://sequelize.org - but it does have a migration framework, which seems like something we should consider.

PR #31, opened 2020-05-07
Add organizations collection

ibeckermayer:
I added an `organizations` collection to the data model and am enforcing that every user has a an `organizationID` field that corresponds to a valid organization.

Added a bunch of tests to both enforce this and as a form of documentation for the frontend devs. Also updated [this Notion doc](https://www.notion.so/covidwatch/Org-Admin-App-Technical-Details-f8a235f8cfb44e1d938c731ccfe621cb) with the new design


jcolla-holla:
I got this running locally and:
- passed all the Jest tests
- ran through the workflow Isaiah set up manually successfully, caught an edge case that Isaiah fixed

May make sense to merge this now so we can proceed further.

PR #34, opened 2020-05-09
Added windows build script

fordacious:
script assumed bash so added windows compatible scripts alongside them.


ibeckermayer:
Are you hoping to work on the project from a windows machine? Currently the work is largely divided out and nobody is using a windows machine. If you want to help out, would you mind connecting in the Covid Watch Slack?

PR #38, opened 2020-05-10
Code validation screen

jcolla-holla:
I wanted to get comfortable using the React Material UI library so added the success/failure messages and a loading graphic.  The loading graphic we can just comment it out for now -- I'm also not sure how long the data processing on the 'verify my code' action will take, it may not be sufficiently long to justify a loading graphic at all.  We'll see.

I git pulled from Kim's in progress branch to make sure the way I implemented the material UI library is the same as her so we don't have possible merge conflicts in the coming day or two.  So there are a handful of files updated in this change that aren't necessarily reflecting the work related to `CodeValidation.js` and `code_validation.scss` where the majority of work happened.

### Success message and graphic loading demo here:

ezgif com-video-to-gif (12)


fordacious:
"I'm also not sure how long the data processing on the 'verify my code' action will take, it may not be sufficiently long to justify a loading graphic at all. We'll see."

If its a standard request over the internet it should at least long enough for a graphic.

jcolla-holla:
@therealnikhil Updated with a new commit to have the Snackbar alert appear just below the NavBar.  See gif below:

ezgif com-video-to-gif (13)

fordacious:
Also just a note, we should integrate this with the pending operation component, as that logic is encapsulated there. Can be done after either is merged.

PR #39, opened 2020-05-10
Added windows build script

fordacious:
I'll be using windows, so added a script to the package.json which is windows compatible


ibeckermayer:
Ok. Can you please download this repository and recreate these as branch merges (rather than fork merges) so they can more easily be tested? I'll add you as a contributor (look for an invite in your email / github notifications)

PR #62, opened 2020-05-11
User authentication state management using mobx state tree

therealnikhil:
You should see the store object in client/src/store/index.js. It's simple now, with just the isAdmin field but can be extended as we continue backend integration with the frontend. Functionality wise, it is the same as before but instead of using the "auth" object we now have a dedicated global storage system in place.


client/src/store/index.js
```diff
@@ -0,0 +1,50 @@
...
+    signIn
+  }
+})
+
+const Store = types
+.model({
+  user: types.maybeNull(User)
+})
+.actions(self => {
+  const signOut = () => {
```
ibeckermayer, line 37: Can we add the actual firebase logout function here?
```javascript
firebase.auth().signOut().then(function() {
  // Sign-out successful.
}).catch(function(error) {
  // An error happened.
});
```

PR #73, opened 2020-05-13
Team managment

wjoe2046:
Hi we fixed all the merge errors between branch dev and branch team-managment. It is now working with the following command:

`npm run start:dev`

Please accept all the incoming changes; please advise if further modifications are needed. 

PS: There are some style issues etc that need to be fixed. @rquartararo plan to work on those in the next few days.


client/src/screens/ManageTeams.js
```diff
@@ -1,21 +1,234 @@
...
+  PreviousPage: forwardRef((props, ref) => <ChevronLeft {...props} ref={ref} />),
+  ResetSearch: forwardRef((props, ref) => <Clear {...props} ref={ref} />),
+  Search: forwardRef((props, ref) => <Search {...props} ref={ref} />),
+  SortArrow: forwardRef((props, ref) => <ArrowUpward {...props} ref={ref} />),
+  ThirdStateCheck: forwardRef((props, ref) => <Remove {...props} ref={ref} />),
+  ViewColumn: forwardRef((props, ref) => <ViewColumn {...props} ref={ref} />),
+}
+//enabling button icons in material table
+
+const useStyles = makeStyles((theme) => ({
```
fordacious, line 50: can styling be moved to sass for better maintainability and comparability with existing styling?

client/src/screens/ManageTeams.js
```diff
@@ -1,21 +1,234 @@
...
+      { name: 'Donald J Trump', role: 1, status: 0 },
+    ],
+  })
 
-const ManageTeamsBase = () => {
   return (
-    <h1>Manage Teams</h1>
-  );
-};
+    <Grid container direction="column" style={{ padding: '3em' }}>
```
fordacious, line 131: same with these styles

client/src/screens/ManageTeams.js
```diff
@@ -1,21 +1,234 @@
...
+        <Button className={classes.saveChanges}>Save Changes</Button>
+      </Grid>
+    </Grid>
+  )
 }
 
-const ManageTeams =  compose(
-  withAuthorization(condition),
-)(ManageTeamsBase);
+// const condition = (authUser) => {
```
fordacious, line 227: can remove

client/ui/Theme.js
```diff
@@ -0,0 +1,36 @@
+import { createMuiTheme } from '@material-ui/core/styles';
+
+export default createMuiTheme({
```
fordacious, line 3: same style comments

therealnikhil:
Couple of notes:
- The table looks nothing like the one on figma
- The table is buggy when I try to order by name or any field
- When you try to view 25, 100 etc. per page, a lot of empty space is shown
- Name should not be editable

client/src/screens/ManageTeams.js
rquartararo: We are using Material UI for the table component, so styling needs to be done a bit differently. Honestly, this is my first time using material UI and it seems a bit limiting as you have to override the default styles if you want to use custom CSS. I'll go over it tonight and see if I can make it better bc I agree it doesn't look clean at all.

PR #74, opened 2020-05-13
Non admin flow complete

### LINKED ISSUE #8: Tab Navigation
If a user is an organization admin, they should have navigation that lets them choose between the permission validation screen, the user management screen, and the organization settings screen. Non-organization admins should only see the permission validation screen and my settings screen (should not be able to change roles here).

### LINKED ISSUE #71: Allow logins for non admins
According to figma v2 flow, non-admins capabilities and restrictions:
- should only be able to view code validations and settings screens.
- should not be able to edit their role in settings

therealnikhil:
Closes #8 , closes #71

### DIFF
--- a/client/src/components/NavBar.js
+++ b/client/src/components/NavBar.js
@@ -102,12 +102,18 @@ const NavBarBase = () => {
         <MenuItem style={linkStyles} onClick={() => onClickMenuItem(0)}>
           Positive Test Validations
         </MenuItem>
-        <MenuItem style={linkStyles} onClick={() => onClickMenuItem(1)}>
-          Manage Members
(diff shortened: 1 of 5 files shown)


therealnikhil:
Use `admin@soylentgreen.com` to test admin, and `user@soylentgreen.com` for non admin

ibeckermayer:
One issue I've found is that if I try to navigate to a blocked route while not logged in, I get a white screen and the console throws an error. The same happens if I navigate to an admin-only route as a non-admin:

Screen Shot 2020-05-14 at 12 21 38 PM

I'll let you decide whether to fix that here or make it a separate issue.

PR #78, opened 2020-05-14
Add member modal

rquartararo:
1. Added a modal for adding new members. 
2. I don't think we have the figma design for the modal so I just put in a simple form styled off the login page. 
3. Just realized we need a modal for the login page as well so I'll start working on that as well.


rquartararo:
@jcolla-holla It should be rendering on the ManageTeams.v1.js, which is the new table that @therealnikhil created.

PR #87, opened 2020-05-15
Added success / failure toast to branding page

fordacious:
Added a success / failure toast which is based on whether the server operation succeeds or fails. Removes itself after 6 seconds.

Componetized this in components/toast.js so it + the save button (components/pendingoperationbutton.js) may be reused on all screens that have data submission.

image
image


ibeckermayer:
Good call, note that there was a big repo restructure which may be the cause of those conflicts

PR #88, opened 2020-05-15
Merge from report-service branch

joshlf:


ibeckermayer:
Deprecated, going with Google's server implementation instead

PR #129, opened 2020-05-21
Context based state manager

### LINKED ISSUE #111: Make mobx react to firestore updates, and make stateful components react to mobx
Currently, the app doesn't have a clear architecture for synchronously responding to `Firestore` changes. The simplest way to set up the problem is through an example:

Say there are two admin users (Alex and Bobby) of the same organization, using the app at the same time. If Alex creates a new user, that user should appear on Bobby's "Manage Members" screen, without him having to refresh. Same goes for if Bobby changes something in the "Account Branding" -- Alex should know about it right away.

Firestore provides us a platform for handling this in the form of [document listening ](https://firebase.google.com/docs/firestore/query-data/listen)(`onSnapshot`). We can take advantage of this by architecting the app such that
- mobx runs listeners, that when triggered update the mobx state
- React components are reactive w.r.t. mobx state

With this architecture, components can write directly to Firestore without needing to worry about updating the global state. Additionally, we get multi-client synchronization for free, since other client instances will be listening for such changes and updating state based on them.

ibeckermayer:
Big time architecture change: This PR removes `mobx` entirely, and replaces the global state manager with a system based on [React Contexts](https://reactjs.org/docs/context.html). The design is largely taken from [this blogpost](https://itnext.io/manage-react-state-without-redux-a1d03403d360).

The motivation behind this proposal is to maintain app state across clients, as described in https://github.com/covid19risk/permission-portal/issues/111. As you can tell from that description, initially I intended to keep `mobx`, but in building this realized that neither it (nor Redux) are up to the task. The problem with both libraries is that they make it arbitrarily complex to attach and detach Firebase document [listeners](https://firebase.google.com/docs/firestore/query-data/listen) based on app state, and update app state based on the listeners. With this design, we can manage our state with all the flexibility of an ordinary React component class, while getting reactive virtual DOM updates for free.

The magic happens primarily in `src/store/index.js`, where `createStore` returns an [HOC](https://reactjs.org/docs/higher-order-components.html) that defines the store. `createStore` is called only once, in `App.js`, ensuring that there is only ever a single store, at the highest level in our component hierarchy. `createStore` takes our App and wraps it in a context [Provider](https://reactjs.org/docs/context.html#contextprovider), and passes this store component's `state` object as a prop:
```javascript
...
render() {
      return (
        <StoreContext.Provider value={this.state}>
          <WrappedComponent {...this.props} />
        </StoreContext.Provider>
      )
    }
```

With the entire app is wrapped by that `Provider`, and its value is set to our store component's `state`, we can now make any other component in our hierarchy subscribe to that `state`'s updates by making it a `Consumer` of the context. The function `withStore` does this for us, so that any component we wrap with it gets a prop called `store` (hence heavy usage of `props.store` throughout the app), that points to our highest level `state` object. You can see that all of our components that make use of global state are now wrapped with `withStore`:
```javascript
const withStore = (WrappedComponent) => {
  return class extends React.Component {
    displayName = 'storeConsumer'
    render() {
      return (
        <StoreContext.Consumer>{(store) => <WrappedComponent store={store} {...this.props} />}</StoreContext.Consumer>
      )
    }
  }
}
```

The rest of the magic is contained in `this.authStateListener`, which is set up when the app loads up to listen for a user logging in or out. When a user logs in, this function goes through and updates the global `state` by grabbing relevant documents from the db, and sets up listeners that wait for changes in relevant documents. When a listener is notified of a change, it updates the global state variable accordingly. 

Importantly, the programmer should (almost?) never change the properties of the global `state` variable directly, except from within the component returned by `createStore` (and should be sure to always call `setState` when doing so). **Data should only flow one way** -- 
- a component makes a call to a function exposed in the global `state` variable (i.e. `updateOrganization`)
- that function makes a call modifying the Firestore database
- a listener is notified of that modification (on all currently running clients), and modifies the global `state` variable accordingly
- any components that were rendering something based on global `state` data (i.e. are wrapped with `withStore` and making use of `props.store`) are automatically re-rendered.

A bit rambly, but hopefully more clarifying than befuddling. Hit me up in Slack if you have any questions.

closes #111


ibeckermayer:
Update: after discussion with @therealnikhil, this is being revised to make use of `mobx`

PR #152, opened 2020-05-23
Refactor GitHub Workflows

colbymorrison:
Merges the frontend and firebase GitHub workflows and adds staging/prod deployment for Firebase.

Now there are only 2 workflows, `build` and `deploy.` `build` is ran on a PR to dev or master and `deploy` is ran on a push to dev or master.

Both workflows build the frontend app and deploy and test firebase on the test infra. `deploy` then deploys to staging or prod based on the branch pushed.

### Question
I just added `NODE_ENV` to the Action's environment for deploying Firebase to staging/prod. Is this correct to make it work? Do we want to do this with different `npm run` commands instead?

### TODO from here
Rename current Firebase key in secrets to `FIREBASE_TOKEN_TEST`
Add prod and staging Firebase keys to secrets


.github/workflows/build.yml
```diff
@@ -1,16 +1,14 @@
-# Install Node dependencies and build app
+# Ensure apps build and tests pass
```
ibeckermayer, line 1: Let's name this file something more accurate, such as `build-and-test.yml`

.github/workflows/build.yml
```diff
@@ -30,3 +28,28 @@ jobs:
           npm run init
           npm run git-info
           npm run build
+
+  test-firebase:
```
ibeckermayer, line 32: Can we break this one down into 3 separate jobs, ie. build, deploy to test infrastructure, and run tests? It looks like GitHub is [dragging its feet](https://github.community/t5/GitHub-Actions/Ability-to-rerun-just-a-single-job-in-a-workflow/td-p/41629) on the obvious feature of being able to re-run specific jobs, but this will prepare us for the (possibly mythical) day they release that feature (for example, if our build succeeds but our test run fails, sometimes we just want to rerun the tests again without waiting for a re-build). Breaking this down will also prepare us for a more complex testing workflow that includes frontend tests and integration tests.

See [here](https://help.github.com/en/actions/configuring-and-managing-workflows/persisting-workflow-data-using-artifacts#passing-data-between-jobs-in-a-workflow) for how to persist data between jobs, and denote that certain jobs are dependent on other jobs.

.github/workflows/deploy.yml
```diff
@@ -53,3 +54,52 @@ jobs:
...
+
+      - name: Install
+        working-directory: permission-portal-cloud-functions/functions
+        run: npm install
+
+      - name: Initialize config files
+        working-directory: permission-portal-cloud-functions/functions
+        run: npm run init        
+
+      - name: Build and Deploy to Firebase Test Infra
```
ibeckermayer, line 72: Redundant, we can assume that if a branch has been merged into master or dev, tests have already run and passed (in the other workflow).

.github/workflows/deploy.yml
```diff
@@ -53,3 +54,52 @@ jobs:
...
+        uses: w9jds/firebase-action@master
+        with:
+          args: deploy --project=test
+        env:
+          FIREBASE_TOKEN: ${{ secrets.FIREBASE_TOKEN }}
+          PROJECT_PATH: permission-portal-cloud-functions/
+
+      - name: Run Tests
```
ibeckermayer, line 80: Also redundant, can be deleted

colbymorrison:
Above issues should be resolved. Building and deploying of firebase has been separated into 2 jobs in both `build-and-test` and `deploy` workflows. The build job uploads the `functions` dir which is then downloaded by the deploy job. Note that we exclude `node_modules` in the upload/download and rerun `npm install` in the deploy job. This is because of the classic node issue of `node_modules` being large (~250mb in this case). When I tried this with the `node_modules` being uploaded and downloaded, the workflow took 45min instead of 11 and caused some permission errors.

PR #169, opened 2020-05-26
Navbar profile img revised

ibeckermayer:
- revised My Settings so that you don't need to click the "save settings" button in order to save your profile image
- added a new collection in our firestore called `userImages` so that the app isn't downloading large amounts of useless image data (which is what would happen if we stored each image blob in the user document itself)
- updated rules to allow for reading/writing of `userImages` collection
- updated the store to handle all this new logic

This is the next iteration of @prabhuinbarajan 's work in this PR (now closed): https://github.com/covid19risk/permission-portal/pull/149

### DIFF
--- a/permission-portal-cloud-functions/firestore.rules
+++ b/permission-portal-cloud-functions/firestore.rules
@@ -47,5 +47,9 @@ service cloud.firestore {
       allow read: if isSignedIn() && isInOrganization(organizationID);
       allow write: if isSignedIn() && isAdmin() && isInOrganization(organizationID);
     }
+
+    match /userImages/{email} {
+	    allow read, write: if isSignedIn() && isSelfOrAdminInOrg(email, resource.data.organizationID);
+    }
   }
 }
--- a/permission-portal-frontend/client/src/components/NavBar.js
+++ b/permission-portal-frontend/client/src/components/NavBar.js
@@ -61,7 +61,10 @@ const NavBarBase = observer((props) => {
         <div className="title">{props.store.user.isAdmin ? ROLES.ADMIN_LABEL : ROLES.NON_ADMIN_LABEL}</div>
       </div>
       <div className="avatar_group avatar_image">
-        <img src={profile} />
+        <img
+          src={props.store.user.imageBlob ? props.store.user.imageBlob : profile}
+          style={{ objectFit: 'cover', display: 'block', margin: 'auto' }}
+        ></img>
       </div>
       <div className="avatar_group separator" />
       <IconButton edge="start" className={classes.menuButton} color="inherit" aria-label="menu" onClick={handleMenu}>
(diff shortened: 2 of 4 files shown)


permission-portal-frontend/client/src/components/NavBar.js
```diff
@@ -61,7 +61,10 @@ const NavBarBase = observer((props) => {
         <div className="title">{props.store.user.isAdmin ? ROLES.ADMIN_LABEL : ROLES.NON_ADMIN_LABEL}</div>
       </div>
       <div className="avatar_group avatar_image">
-        <img src={profile} />
+        <img
+          src={props.store.user.imageBlob ? props.store.user.imageBlob : profile}
+          style={{ objectFit: 'cover', display: 'block', margin: 'auto' }}
```
fordacious, line 66: can be in sass file?

permission-portal-frontend/client/src/screens/Settings.js
```diff
@@ -152,14 +151,16 @@ const SettingsBase = observer((props) => {
...
     e.preventDefault()
-    try {
-      await props.store.user.update({ ...state })
-      console.log('user data saved successfully')
-      setToastInfo({ open: true, success: true, msg: 'Settings Saved Successfully' })
-    } catch (err) {
-      console.log(err)
-      setToastInfo({ open: true, success: false, msg: 'Failed' })
-    }
+    console.log('Deprecated, TODO: delete this functionality')
```
fordacious, line 154: can it just be deleted now?

ibeckermayer: Yeah, good point.

permission-portal-frontend/client/src/components/NavBar.js
ibeckermayer: Probably, though I see other styles scattered about this file and others. I'm going to leave this for now, though a useful task might be to refactor the code to conform to some styling paradigm.

permission-portal-cloud-functions/firestore.rules
```diff
@@ -47,5 +47,9 @@ service cloud.firestore {
       allow read: if isSignedIn() && isInOrganization(organizationID);
       allow write: if isSignedIn() && isAdmin() && isInOrganization(organizationID);
     }
+
+    match /userImages/{email} {
+	    allow read, write: if isSignedIn() && isSelfOrAdminInOrg(email, resource.data.organizationID);
+    }
```
prabhuinbarajan, line 53: i dont know if an admin is ever required to modify the profile images. it is typical for this rule to be isSelf only

permission-portal-frontend/client/src/components/NavBar.js
prabhuinbarajan: i think that is a good call. overall how styles are handled in app is fragmented. @fordacious  - when we refactor, we should move these styles in sass , import and do the usual react way

ibeckermayer:
closes https://github.com/covid19risk/permission-portal/issues/150

permission-portal-cloud-functions/firestore.rules
ibeckermayer: Sorry, missed this before I merged. You're right, admin doesn't need to be able to modify, will fix in a different PR

PR #178, opened 2020-05-28
Center Modals

### LINKED ISSUE #173: "Forgot Password?" Modal is too high on smaller screens
The modal that pops up when I click forgot pwd on my macbook hugs the top of the screen, which looks kind of jank:
Screen Shot 2020-05-26 at 7 18 23 PM

Let's make it centered for all screen heights

colbymorrison:
Closes #173 

Figured this should apply to the Add Member Modal as well. Created `modal.scss` for common styling between all modals including this change, which always centers them. 

Also moved the Forgot Password modal to the `components` dir, because I think it should live there and not `screens`


colbymorrison:
> Good idea to encapsulate common styling, though the add member modal is getting cut off at the top for me:

This seems to be an issue on `dev` as well. I added a couple height CSS params to the add member modal in the most recent commit, but past a certain screen height this is still an issue

Screen Capture_select-area_20200529005600

The input box/button heights are fixed in `add_member_modal.scss` so at a certain point we can't make the modal any smaller without things looking silly. So I think a more significant styling rewrite would be necessary.

PR #186, opened 2020-05-29
Removing "self" from the Manage Members table

ibeckermayer:
With this change, when an admin is signed in they won't see their own entry in the Manager Members table, since they can modify their own settings on their Settings page.


aidan-fitz:
If only `!=` operators were supported :frowning_face:

PR #189, opened 2020-05-29
Delete user backend

ibeckermayer:
One half of https://github.com/covid19risk/permission-portal/issues/90


permission-portal-cloud-functions/functions/src/index.ts
```diff
@@ -51,38 +51,44 @@ function doesOrganizationExist(organizationID: string): Promise<boolean> {
...
+  return auth
     .getUser(uid)
     .then((userRecord) => {
       const email = userRecord.email;
-      auth
+      return auth
         .deleteUser(uid)
         .then(() => {
-          usersCollectionDeleteUser(email ? email : '');
+          return usersCollectionDeleteUser(email ? email : '');
```
fordacious, line 82: is blank email valid? I suppose it'll hit the user in org check.

permission-portal-cloud-functions/functions/src/index.ts
```diff
@@ -51,38 +51,44 @@ function doesOrganizationExist(organizationID: string): Promise<boolean> {
...
 function usersCollectionDeleteUser(email: string) {
-  db.collection('users')
+  return db
+    .collection('users')
     .doc(email)
     .get()
     .then((user) => {
       if (user.exists) {
-        user.ref.delete().catch((err) => {
+        return user.ref.delete().catch((err) => {
```
fordacious, line 60: same here. Does the catch bubble out?

permission-portal-cloud-functions/functions/src/index.ts
```diff
@@ -267,6 +292,37 @@ export const createUser = functions.https.onCall((newUser, context) => {
...
+    isUserInCallersOrganizationGuard(data.email, context)
+      .then(() => {
+        auth
+          .getUserByEmail(data.email)
+          .then((userRecord) => {
+            deleteUserByUid(userRecord.uid)
+              .then(() => {
+                resolve(`Successfully deleted user ${data.email}`);
+              })
+              .catch((err) => {
```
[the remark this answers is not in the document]
ibeckermayer: Afaik the exceptions technically do bubble out (if you don't handle them they get thrown). The default `tslint` rule that's enforcing this endless chain of catches is
```json
// Functions must either be handled directly (e.g. with a catch() handler) or returned to another function.
    // This is a major source of errors in Cloud Functions and the team strongly recommends leaving this rule on.
    "no-floating-promises": true,
```

Because the comment seems to so heavily recommend it, I've just been going with it.

permission-portal-cloud-functions/functions/src/index.ts
ibeckermayer: The blank email will cause `usersCollectionDeleteUser` to throw an error when we try to get a document indexed by `''`. We could definitely do some more elegant error checking / messages but I made the judgement call that this was good enough for now, if we start hitting this then I'll add some fancier checks.

Issue #181, opened 2020-05-28
Password Reset Successful Toast is not dismissible


ibeckermayer:
The "Password Reset Successful" toast on the "My Settings" page isn't dismissible, and blocks the user from opening the menu. (Presumably the failure toast causes a similar problem)

aidan-fitz:
How should the toast behave? Should it have a dismiss button (like an X) or dismiss when clicked/automatically?

colbymorrison:
@aidan-fitz I added a toast in my previous PR so I got a feel for how they work. The toast calls its onClose method after 3 seconds. I think the bug here is that the onClose method of the toast is not setting the correct state variable to false.

PR #190, opened 2020-05-31
Add login failure toast

### LINKED ISSUE #155: login failure toast

colbymorrison:
Closes #155

### DIFF
--- a/permission-portal-frontend/client/src/screens/Login.js
+++ b/permission-portal-frontend/client/src/screens/Login.js
@@ -18,7 +18,7 @@ const INITIAL_STATE = {
   redirect: false,
   showPassModal: false,
   showToast: false,
-  toastMessage: 'Please enter a valid email and password',
+  toastMessage: 'Error logging in, email or password may be invalid',
 }
 
 const SignInFormBase = observer(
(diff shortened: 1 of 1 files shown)


permission-portal-frontend/client/src/screens/Login.js
```diff
@@ -85,6 +88,12 @@ const SignInFormBase = observer(
           </div>
           <ForgotPasswordModal hidden={!this.state.showPassModal} onClose={this.hideModal} />
         </div>
+        <Toast
+          open={this.state.showToast}
+          onClose={() => this.setState({ showToast: false })}
```
aidan-fitz, line 93: Should we make this the default behavior of Toasts? This would resolve #181.

ibeckermayer: I don't think that quite makes sense. The `Toast` would need to somehow get `this` passed in order to `setState` for its parent component.

ibeckermayer: (Feel free to prove me wrong in another PR, though, perhaps I'm missing something)

Issue #185, opened 2020-05-29
Fix uncaught error on logout


ibeckermayer:
On logout, there is an uncaught error (for both admins and non-admins)

```
Uncaught Error in onSnapshot: FirebaseError: Missing or insufficient permissions.
    at new n (http://localhost:8080/dist/app.js:21541:23)
    at t.ni (http://localhost:8080/dist/app.js:24693:16)
    at t.pi (http://localhost:8080/dist/app.js:24852:195)
    at n.onMessage (http://localhost:8080/dist/app.js:31351:33)
    at http://localhost:8080/dist/app.js:31304:26
    at http://localhost:8080/dist/app.js:31335:37
    at http://localhost:8080/dist/app.js:29265:31
```

PR #196, opened 2020-06-03
Remove unnecessary code duplication

ibeckermayer:
Remove all `.get()`'s in the `createStore` constructor in favor of only calling `.onSnapshot()`, which functions as a `.get()` when its first called.

Closes https://github.com/covid19risk/permission-portal/issues/195

### DIFF
--- a/permission-portal-frontend/client/src/store/index.js
+++ b/permission-portal-frontend/client/src/store/index.js
@@ -15,95 +15,62 @@ const createStore = (WrappedComponent) => {
       this.authStateListener = auth.onAuthStateChanged(async (user) => {
         if (user) {
           console.log('User signed in')
-          // signed in, get user's document from the db
-          const userDocumentSnapshot = await db.collection('users').doc(user.email).get()
-          // update the store with data from this document and set isSignedIn to true
-          rootStore.user.__update({ ...userDocumentSnapshot.data(), email: userDocumentSnapshot.id, isSignedIn: true })
-          // set up a listener to respond to current user's document changes
+          // get user's document from the db and at the same time set up a listener to respond to document changes
           if (this.userDocumentListener === null) {
             this.userDocumentListener = db
               .collection('users')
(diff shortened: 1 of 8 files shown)


permission-portal-frontend/client/src/store/index.js
```diff
@@ -15,95 +15,62 @@ const createStore = (WrappedComponent) => {
...
+                        id: updatedOrganizationDocumentSnapshot.id,
+                      })
+                      if (rootStore.user.isAdmin && this.organizationMembersListener === null) {
+                        this.organizationMembersListener = db
+                          .collection('users')
+                          .where('organizationID', '==', rootStore.organization.id)
+                          .onSnapshot((updatedUsersSnapshot) => {
+                            rootStore.organization.__setMembers(
+                              // See https://stackoverflow.com/a/24806827
+                              updatedUsersSnapshot.docs.reduce((result, userDoc) => {
```
aidan-fitz, line 49: Could be more efficient to listen for changes here (for example, see [this code sample](https://firebase.google.com/docs/firestore/query-data/listen#view_changes_between_snapshots))

ibeckermayer: Almost certainly is more efficient. Added an issue for it, though we should wait until pagination has been sorted out to implement: https://github.com/covid19risk/permission-portal/issues/199

colbymorrison:
This fixes #185 as well! I believe with lines 87-90. In `dev` we didn't unsubscribe from `userImageListener` and now we do!

ibeckermayer:
> This Closes #185 as well! I believe with lines 87-90. In `dev` we didn't unsubscribe from `userImageListener` and now we do!

Ah cool, good catch

ibeckermayer:
> This Closes #185 as well! I believe with lines 87-90. In `dev` we didn't unsubscribe from `userImageListener` and now we do!

Ah cool, good catch

PR #200, opened 2020-06-04
Fix Mobx Error on user status change

### LINKED ISSUE #197: Changing a user's status (Active/Disabled) causes a mobx error
When an admin toggles a user between Active/Disabled, we get an error along the lines of
```
Uncaught Error: [mobx-state-tree] Cannot modify 'AnonymousModel@/organization/members/15', the object is protected and can only be modified by using an action.
    at fail$1 (mobx-state-tree.module.js:3103)
    at ObjectNode../node_modules/mobx-state-tree/dist/mobx-state-tree.module.js.ObjectNode.assertWritable (mobx-state-tree.module.js:1541)
    at Array../node_modules/mobx-state-tree/dist/mobx-state-tree.module.js.ModelType.willChange (mobx-state-tree.module.js:4802)
    at interceptChange (mobx.module.js:2992)
    at ObservableObjectAdministration../node_modules/mobx/lib/mobx.module.js.ObservableObjectAdministration.write (mobx.module.js:4006)
    at Object.set [as disabled] (mobx.module.js:4239)
    at onChange (ManageTeams.js:157)
    at HTMLUnknownElement.callCallback (react-dom.development.js:188)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:237)
    at invokeGuardedCallback (react-dom.development.js:292)
```
It looks to me as though somewhere we're trying to set a mobx model's field directly, rather than through an action. Note that this error doesn't appear for other UI actions that cause the mobx-state-tree's list of users to change (such as adding/deleting/changing-roles).

colbymorrison:
Closes #197 

The error was caused by line 156, which was directly editing the model field. We don't need that line or the line after it, because `updateUserByEmail` invokes the `organizationMembersListener` (in `store/index.js`) which updates the user model. The `select` component is reactive, so it will update the `className` automatically when the user model is updated. 

I tested by activating and deactivating my user, and everything works as expected.

### DIFF
--- a/permission-portal-frontend/client/src/screens/ManageTeams.js
+++ b/permission-portal-frontend/client/src/screens/ManageTeams.js
@@ -153,8 +153,6 @@ const ManageTeamsBase = observer((props) => {
                       value={!data.disabled ? 'active' : 'deactivated'}
                       onChange={(e) => {
                         props.store.updateUserByEmail(data.email, { disabled: e.target.value == 'deactivated' })
-                        data.disabled = e.target.value == 'deactivated'
-                        e.target.className = !data.disabled ? 'active' : 'inactive'
                       }}
                       aria-labelledby="status-header"
                     >


permission-portal-frontend/client/src/screens/ManageTeams.js
```diff
@@ -153,8 +153,6 @@ const ManageTeamsBase = observer((props) => {
                       value={!data.disabled ? 'active' : 'deactivated'}
                       onChange={(e) => {
                         props.store.updateUserByEmail(data.email, { disabled: e.target.value == 'deactivated' })
-                        data.disabled = e.target.value == 'deactivated'
```
fordacious, line 156: does the styling correctly update without this?

colbymorrison: Yes, the text is still bold when active and not bold when deactivated

PR #202, opened 2020-06-04
Order members by last name, then by first name

### LINKED ISSUE #191: Default sorting for member table
As I discussed on #182, the member table on the Manage Members page should be sorted in a default order, e.g. last name or email address.

aidan-fitz:
Closes #191

This sorts entries in the member table by last name, then by first name.

### DIFF
--- a/permission-portal-frontend/client/src/store/index.js
+++ b/permission-portal-frontend/client/src/store/index.js
@@ -43,8 +43,8 @@ const createStore = (WrappedComponent) => {
                         this.organizationMembersListener = db
                           .collection('users')
                           .where('organizationID', '==', rootStore.organization.id)
-                          // .orderBy('lastName')
-                          // .orderBy('firstName')
+                          .orderBy('lastName')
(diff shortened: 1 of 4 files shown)


permission-portal-frontend/client/src/store/index.js
```diff
@@ -43,6 +43,8 @@ const createStore = (WrappedComponent) => {
                         this.organizationMembersListener = db
                           .collection('users')
                           .where('organizationID', '==', rootStore.organization.id)
+                          // .orderBy('lastName')
+                          // .orderBy('firstName')
```
aidan-fitz, line 47: Commented out because we have to [create an index][index] first.

[index]: https://firebase.google.com/docs/firestore/query-data/indexing

aidan-fitz:
The order is case-sensitive, as Firestore doesn't support case-insensitive sorting. We would have to do something like [this](https://stackoverflow.com/a/48098215) if we wanted to support case-insensitive sorting.

PR #206, opened 2020-06-06
Laford encapsulate logging

fordacious:
Re-initiating and fixing issues.


fordacious:
Once logging is encapsulated, ALL logs in the system can be either turned off or redirected to a custom logging system.

fordacious:
e.g. logs can be automatically turned on for non prod builds and removed otherwise using a build system configuration.

fordacious:
Using this we can safely add as many logs as we want without worrying about it getting to prod

PR #224, opened 2020-06-08
Update 404 page UI

jcolla-holla:
404 page UI update according to v3 Figma complete.  

Added some the logic to `footer.js` as we now have three footer render possibilities:
1) the login screen footer design
2) the 404 screen footer design
3) every other screen footer design

The second option above wasn't a concern before this PR so turnery statements got the job done.  I simply updated from turnery logic to a series of conditionals.  Open to feedback on this if people have strong feelings about it.

Given that this is a desktop based app (no mobile use case), the svg images of the doctors currently have a `min-width` which creates empty white space not covered by the linear-gradient background if the window is a small width, but shouldn't be a concern.

## Visual preview:

Screen Shot 2020-06-16 at 11 04 32 PM


frontend/client/Styles/screens/404.scss
```diff
@@ -1,17 +1,40 @@
 #not-found {
     display: flex;
+    justify-content: space-evenly;
+    // background-image: url('../../assets/fade-background-404.svg');
```
ibeckermayer, line 4: Can any of these comments be deleted?

jcolla-holla: Yes, absolutely.  Left those there by accident when experimenting w `linear-gradient` instead of svg file background.

PR #229, opened 2020-06-09
Manage users pagination

ibeckermayer:
#### Note
The most meaningful of the changes in this PR are in `./permission-portal-frontend/client/src/store/index.js`, `./permission-portal-frontend/client/src/store/model.js`, and `./permission-portal-frontend/client/src/screens/ManageTeams.js`. The reason so many other files are changed in this PR is that an architecture change required a minor change in all the components that made use of `props.store` (explained below).

#### Changes
Based on [this Slack thread](https://covid-watch.slack.com/archives/C013MKUMZK3/p1591399870300700), we are to handling pagination by only allowing the user to go to the next page or previous page, rather than jump to arbitrary page numbers, which Firebase isn't well designed to support.

I realized that in order to handle pagination, we need to be able to swap out listeners in our master ReactContext component (the component that wraps the passed component in `createStore`, in `store/index.js`, which I will refer to as the "master component").  Specifically, whenever we change pages, we need to be able to swap `this.organizationMembersListener` (now renamed to `this.__pageOfMembersListener`) out for a new query listener, representing our new page.

Previously we were only passing the mobx-state-tree `rootStore` object as our store:
```javascript
render() {
      return (
        <RootStoreContext.Provider value={rootStore}>
          <WrappedComponent {...this.props} />
        </RootStoreContext.Provider>
      )
    }
```
However with this design, we had no way to make calls to the master component from anywhere besides its constructor. Specifically, I needed a way for the `ManageTeams.js` react component to call to the master component and tell it to swap out its current page of members listeners (`this.__pageOfMembersListener`) for the next/prev page. To solve this, I changed our `RootStoreContext.Provider` value to the master component:
```javascript
render() {
      return (
        <RootStoreContext.Provider value={this}>
          <WrappedComponent {...this.props} />
        </RootStoreContext.Provider>
      )
    }
```
and added the `rootStore` (the object we were previously accessing via `props.store`) as a field on the master component in it's constructor:
```javascript
this.data = rootStore
```
and then replaced all occurrences of `props.store` with `props.store.data`. This gives us the best of both worlds, in that we have access to a fully functional javascript object as our global state manage / data store, and still get to take advantage of all of mobx-state-tree's fancy smart re-rendering.

It also means that all of the master components' fields and methods are now exposed to every sub-component that makes use of `props.store`. By convention, I've appended "`__`" to the fields and methods that are meant to be private.

The pagination magic happens in `nextPageOfMembers()`, `previousPageOfMembers()`, and `__updatePageOfMembersOnSnapshot()`.

#### Manual Test
The Initech org on the dev infrastructure now has 5000 members, uploaded using the updated `permission-portal-cloud-functions/functions/reset-infrastructure.js` script. Login as `admin@initech.com` and you can page through members.

closes https://github.com/covid19risk/permission-portal/issues/182

#### Update
Realized some bugs in working on https://github.com/covid19risk/permission-portal/issues/234. Fixed them and am simply adding https://github.com/covid19risk/permission-portal/issues/234 to this PR.
closes https://github.com/covid19risk/permission-portal/issues/234

#### Further Update
In the spirit of restructuring, I removed the redundant "permission-portal-" from the top level directories, and changed the github workflows accordingly


fordacious:
This seems quite complex and limited (can't jump page to page). What is the cost of doing pagination client side?

aidan-fitz:
Why can't we have page numbers, at least the ones for the current, previous, and next pages? We don't want to show every page such that the page links stretch across the screen, like it does before this PR:

The Initech org before this PR

On the other hand, without any numbers, the navigation widget looks like this:

Without page numbers

PR #238, opened 2020-06-10
Track cloud functions "env vars"

### LINKED ISSUE #215: Version control cloud functions "environment variables" and add to GitHub actions
Our cloud functions currently make use variables set via Firebase's [environment configuration](https://firebase.google.com/docs/functions/config-env) system. At the time of writing, we are using `functions.config().client.url` and `functions.config().sendgrid.key`.

We should design and implement a system for version controlling such variables. The system needs to be variable between projects (dev/test/staging/prod), not divulge secrets (such as our SendrGrid api key), and be callable from our GitHub actions so that we can update these variables automatically during deployment.

colbymorrison:
Closes #215.

First attempt at solving this issue, comments appreciated.

My idea is:

- For non-secret env vars (currently just `client.url`), put them in `config/functions.config.[project].json` I created these json files for dev/test/prod/staging.

- For secret env vars (currently `sendgrid.api_key`) add them as GitHub secrets. We may also want to add the key but not the value to the config file mentioned above, so we can note in our Version Control that that env-var should be set. 

Created the file `permission-portal-cloud-functions/functions/set-config.js` and set `npm run config:set` to run that file.

The file loads in the appropriate json file from `config` containing the non-secret env vars (call this the local config). It then adds to the local config object the secret env-vars from environment variables set before running the script.

It then checks the current cloud-functions environment, if this is the same as the local config, it stops.

If there are changes, it sets the new env-vars using a method from [this firebase issue](https://github.com/firebase/firebase-tools/issues/406) discussing this situation.

So, to run this locally we can do 
```
export SENDGRID_API_KEY=key
export FIREBASE_TOKEN=token 
export NODE_ENV=dev
npm run config:set
```

### Updates
- Config functions are now `.js` and contain both secret and non-secret values as per comment below

- GitHub actions have been updated to run this on build and deploy.


colbymorrison:
> > So, to run this locally we can do
> > ```
> > export SENDGRID_API_KEY=key
> > export FIREBASE_TOKEN=token 
> > export NODE_ENV=dev
> > npm run config:set
> > ```
> 
> Is `FIREBASE_TOKEN` necessary here? My understanding is that this is set via GitHub to be used in our Actions, but it shouldn't be necessary here. But perhaps I'm missing something.

You're right, it isn't necessary if you're signed in to the Firbease CLI, but we will use it with the GitHub actions

colbymorrison:
> Great work, I like where this is headed. My only qualm with this design is that these `functions.config.*.json` files don't accurately represent what the remote config actually should be -- they only give a partial representation (the non-secret aspects of our config), and then we inject our secrets in later in `set-config.js`.
> 
> I think we can make these more accurate by changing them to `.js` files, building the config object in the file and including the secrets as env variables, and then exporting the object for import by `set-config.js`.
> 
> Ex:
> 
> ```js
> var config = {
>     "client": {
>         "url": "http://localhost:8080/"
>     },
>     "sendgrid": {
>         "key": process.env.SENDGRID_API_KEY
>     }
> }
> 
> export default config
> ```

Agreed! I was thinking this as well last night. Will update...

PR #241, opened 2020-06-10
Make first time users update password

### LINKED ISSUE #214: Force first time users to update their password
Create a new boolean field in our user model named `isFirstTimeUser`. On login, if the user is a first time user, they should not be able to do anything until they successfully change their password via the Change Password modal, as shown in the v3 Figma frame `New Member / Change Password`.

This should make use of the client lib's [updatePassword](https://firebase.google.com/docs/auth/web/manage-users#set_a_users_password) function

aidan-fitz:
Closes #214 

Prompt the user to set a new password when they log in if the `isFirstTimeUser` flag is set.

To test:
1. Make sure that the cloud functions from this PR are on the server.
2. Create a new user. It will have `isFirstTimeUser: true`.
3. Log in as the new user from the home page.
4. In the `ChangePasswordModal`, enter a new password and click "Save".

Notes:
- Firebase requires passwords to be 6 characters long and throws `auth/weak-password` if the submitted password is not strong enough. We enforce this requirement on the client side by disabling the submit button unless it is met.
- Firebase throws `auth/requires-recent-login` if the user logged in too long before a call to [`updatePassword()`][updatePassword]. We handle this by prompting the user to [reauthenticate] using their current password and retry the operation.

[updatePassword]: https://firebase.google.com/docs/reference/js/firebase.User#updatepassword
[reauthenticate]: https://firebase.google.com/docs/reference/js/firebase.User#reauthenticatewithcredential


permission-portal-frontend/client/src/components/ChangePasswordModal.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react'
+import Modal from '../components/Modal'
+import PendingOperationButton from '../components/PendingOperationButton'
+
+class ChangePasswordModal extends React.Component {
```
fordacious, line 5: This boilerplate code is different from other components we have.

permission-portal-frontend/client/Styles/screens/change_password_modal.scss
```diff
@@ -0,0 +1,83 @@
+.change-password-modal-container {
```
fordacious, line 1: Don't need to do it now but it seems like we're defining modals multiple times atleast style wise

aidan-fitz: I agree, we should refactor the CSS a bit. It's annoying to have to re-engineer the styling in the mockup every time we create a new component.

ibeckermayer: Can you please create an issue for this?

permission-portal-frontend/client/src/components/ChangePasswordModal.js
ibeckermayer: @fordacious what do you mean by that? As in its a class component rather than a function component?

aidan-fitz:
I just implemented something like this.

permission-portal-frontend/client/Styles/screens/change_password_modal.scss
aidan-fitz: See #253

PR #248, opened 2020-06-11
Adding initial security policy

whaber:
This will be replaced with something better in the future, but is ok to start with this.


whaber:
Ref:

- https://github.com/covid19risk/covidwatch-ios-en/blob/master/SECURITY.md
- https://github.com/covid19risk/covidwatch-ios-en/blob/master/SECURITY.md
- https://covid-watch.slack.com/archives/G0139ER4Y68/p1591639799063100?thread_ts=1591381130.061000&cid=G0139ER4Y68

Issue #253, opened 2020-06-12
Refactor SCSS


aidan-fitz:
We should create reusable SCSS styles that we can use whenever we want to create a new component. To avoid duplication, style rules should be as high up in the component hierarchy as possible: for example, a style rule that applies to all modals should be applied to `Modal`, not `ChangePasswordModal`, `AddMemberModal`, etc.

Follow-up to #241

aidan-fitz:
Update: Just move redundant style rules into `modal-container` and `button`

PR #255, opened 2020-06-13
Update styles to match v3 figma

### LINKED ISSUE #225: Header and body margins
According to the v3 mockup:

- The left and right margins of the page header and body should be between 98px and 100px. All elements on the page should be aligned with the left and right margins.
- There should be 127px between the menu hamburger icon and the right side of the screen.
- Toasts should still stretch across the entire width of the screen.

Example - user settings page:
Screenshot from 2020-06-08 21-58-32

colbymorrison:
Closes #225 

### Changes
- Updates left and right margin on all screens to be `98px` 
- Update navbar to match left margin of screen and add some more right margin to be consistent with the Figma
- Refactor the scss files a bit, moves the `module-container` class to `application.scss` since it is used by all components. More work here needs to be done here in terms of refactoring the Style files, especially those in the `Styles/components` directory as I didn't look at those in this PR, but that doesn't seem pressing for v3 and is noted in #253


colbymorrison:
> On the login page, the form almost aligns with the "Powered by Covid Watch" logo on medium-width screens, but becomes misaligned on wider screens. The logo and form should be right-aligned with each other.

Just pushed a commit to fix this :)

colbymorrison:
> On the settings page, there's extra whitespace to the right of the body content. The form elements on the right side should be aligned with the hamburger icon.
> 
They're not aligning with the hamburger because the input boxes have a `75%` width on them, with `100%` width they overlap. So this would require conditional styling on the nav bar margins based on the page, which doesn't seem worth it imo. There there may be an easier way, but I don't see it, I'm not really that experienced with CSS.

> Also, the first column of text boxes and the text "Changes are automatically saved" should be aligned. They seem to "switch places" when you resize the window.

Yeah, the settings page was giving me hell because of the way it's setup. The top bar with the header and "Changes are automatically saved" are styled with normal CSS. Then the rest of it is a [Material UI Grid](https://material-ui.com/components/grid/), which spaces the 3 columns out over a 12 column grid. So I couldn't figure out a way to get the "Changes are automatically saved" and the second grid item to align as the screen size changes, though I did try for a while before giving up lol. Feel free to check it out if you want, though this doesn't seem too high prio.

aidan-fitz:
Have you tried making the top bar part of the grid?

colbymorrison:
Yes I tried that, but that messed up the vertical spacing on the grid. I could give it one more shot though

PR #256, opened 2020-06-13
Reset Password (settings page)

### LINKED ISSUE #244: Reset Password
See flow with the same title in [Figma v3](https://www.figma.com/file/UObqudM50aROl5yoCmFqT0/Covid-Watch---Portal-v3-(dev)?node-id=0%3A1) for reference.

One tricky aspect of this might be verifying the user's old password. We should at first try simply calling `signInWithEmailAndPassword` as we do in the `Login.js` page, but this will likely trigger our global login listener which may or may not cause some laggy-ness / general mayhem. If that's not sufficient, we can create a cloud function to verify the old password.

Password can be reset via the [updatePassword](https://firebase.google.com/docs/auth/web/manage-users#set_a_users_password) method

colbymorrison:
Closes #244 

Adds `ResetPasswordModal` which is very similar to `ChangePasswordModal` (thanks @aidan-fitz !). After #243 is finished, we can abstract out what is the same between the 3 modals into a common base component. 

### Note
Changes to the `change image modal` component in `Settings` will be overrideen by #246, I just changed them here to avoid errors.

### DIFF
--- a/frontend/client/src/components/ResetPasswordModal.js
+++ b/frontend/client/src/components/ResetPasswordModal.js
@@ -17,7 +17,8 @@ class ResetPasswordModalBase extends React.Component {
       passwordIsValid: false,
       successful: false,
       message: '',
-      formHasBeenEdited: false,
+      newPasswordHasBeenEdited: false,
+      currentPasswordHasBeenEdited: false,
       confirmPasswordHasBeenEdited: false,
     }
     this.onChange = this.onChange.bind(this)
@@ -33,13 +34,13 @@ class ResetPasswordModalBase extends React.Component {
     let fieldContent = event.target.value
     // Serialize updates
     this.setState((state) => {
-      let newState = {
-        formHasBeenEdited: true,
-      }
+      let newState = {}
       if (fieldName === 'current-password') {
         newState.currentPassword = fieldContent
+        newState.currentPasswordHasBeenEdited = true
       }
       if (fieldName === 'new-password') {
+        newState.newPasswordHasBeenEdited = true
         newState.password = fieldContent
         newState.passwordIsValid = newState.password && newState.password.length >= 6
         newState.passwordsMatch = newState.password === state.confirmPassword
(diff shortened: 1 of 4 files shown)


frontend/client/src/components/ResetPasswordModal.js
```diff
@@ -0,0 +1,161 @@
...
+    super(props)
+    this.state = {
+      currentPassword: '',
+      password: '',
+      confirmPassword: '',
+      passwordsMatch: true,
+      passwordIsValid: false,
+      successful: false,
+      message: '',
+      formHasBeenEdited: false,
```
aidan-fitz, line 20: I would rename this to `newPasswordHasBeenEdited` and add another state variable `currentPasswordHasBeenEdited` to indicate whether the user has changed the `current-password` field, so we can also avoid displaying an error message before the field is changed.

frontend/client/src/components/ResetPasswordModal.js
```diff
@@ -0,0 +1,161 @@
...
+
+  handleFirebaseError(err) {
+    Logging.error(err)
+    if (err.code === 'auth/wrong-password') {
+      this.props.onFailure('Current password incorrect, please try again')
+    } else if (err.code === 'auth/too-many-requests') {
+      this.props.onFailure('Too many incorrect attempts. Please try again later')
+    } else {
+      this.props.onFailure('Failed to reset password, please try again')
+    }
```
aidan-fitz, line 64: I like this addition to the error handling method! We should add it to `ChangePasswordModal` as well in case of any unexpected errors there.

frontend/client/src/components/ResetPasswordModal.js
```diff
@@ -0,0 +1,161 @@
...
+    this.setState({ currentPassword: '', confirmPassword: '', password: '' })
```
aidan-fitz, line 65: Why do you think that the form fields should be blanked when an error occurs? I think this is inconvenient to the user; I'd rather they be able to edit what they have already typed.
```suggestion
```

colbymorrison: I did this because if, say, the user puts an incorrect current password we want them to enter it again. But I agree with you that shouldn't clear everything, so I'll remove this and just clear the `currrentPassword` field on the wrong-password error.

aidan-fitz: Sounds good!

frontend/client/src/components/ResetPasswordModal.js
```diff
@@ -0,0 +1,164 @@
...
+
+  handleFirebaseError(err) {
+    Logging.error(err)
+    if (err.code === 'auth/wrong-password') {
+      this.props.onFailure('Current password incorrect, please try again')
+      this.setState({ currentPassword: '', currentPasswordHasBeenEdited: false })
+    } else if (err.code === 'auth/too-many-requests') {
+      this.props.onFailure('Too many incorrect attempts. Please try again later')
+    } else {
+      this.props.onFailure('Failed to reset password, please try again')
```
aidan-fitz, line 65: Nit: I would change these error messages to have periods at the end and avoid comma splices.

PR #259, opened 2020-06-15
Confirm role change

lneild:
I added a confirmation alert when promoting and demoting a user from contributor to admin


lneild:
Thanks for merging it!

PR #264, opened 2020-06-16
Password recovery

ibeckermayer:
## Password Recovery
Big PR for a tricky feature. The goal here was to make it so that when a user requests a password reset (by clicking forgot password), they're sent an email with a link which they click, which automatically brings them to our bespoke password reset interface.

#### Backend
Added a new endpoint ("callable") named `initiatePasswordRecovery`. This endpoint consumes an email address, updates the associated user document to have a field called `passwordResetRequested` set to `true`, and sends a password recovery email with an [email sign-in link](https://firebase.google.com/docs/auth/admin/email-action-links#generate_email_link_for_sign-in).

#### Frontend
The bulk of the complexity in this feature comes in the frontend state management. The naive implementation of this feature is to pop up the `ChangePasswordModal` for any user that's signed in and has `props.store.data.user.passwordResetRequested === true`. However the problem with this implementation is that it opens the legitimate user up to getting locked out of their account, by a malicious actor who simply incessantly requests password resets with the legitimate user's email address. By necessity, `initiatePasswordRecovery` can be called by unauthenticated users.

To account for this, we instead check for that both `props.store.data.user.passwordResetRequested === true` and that the user signed in via the email link before popping up the `ChangePasswordModal`. This ensures that the user actually actively clicked the link in the password recovery email, which we're taking to mean that they indeed do want a password reset. If the password reset was a mistake, or requested by another person, they can simply delete the email and continue signing in to the app as normal.

See `trySignInWithEmailLink` in `Login.js` as a jumping off point for how we handle this.

Other substantial changes are in `ChangePasswordModal.js` and the `store`. I moved all the direct firebase calls from `ChangePasswordModal.js` into functions in the `store` (we should stick to this paradigm as much as possible in general). I also eliminated some of the complexity associated with error handling -- now, if we encounter a Firebase error we just ask the user to retry <generic-password-change-flow>. This eliminates the difficulty of testing states associated with specific errors (like `'auth/requires-recent-login'`). It additionally makes it much simpler to use this same `ChangePasswordModal.js` component for all of the different password changes (firstTimeUser/recovery/change), without having to design and maintain tricky component state logic.

closes https://github.com/covid19risk/permission-portal/issues/243


frontend/client/src/components/ChangePasswordModal.js
```diff
@@ -58,152 +49,108 @@ class ChangePasswordModalBase extends React.Component {
...
-  updatePasswordAndClose(user) {
-    // 1. Change user password to state.password
-    const updatePwd = user.updatePassword(this.state.password)
-    // 2. Set user.isFirstTimeUser to false
-    const setFirstTimeUser = updatePwd.then(() => {
-      this.props.store.updateUser({ isFirstTimeUser: false })
-    }, this.handleFirebaseError)
-    // 3. Close the modal and pop the toast
-    return setFirstTimeUser.then(() => {
+  async tryUpdatePassword() {
```
aidan-fitz, line 70: I love that you changed this to async/await. :rocket:

ibeckermayer:
> Walking through the workflow at a high level makes sense and no red flags jump out at me.
> 
> I did, however, attempt to reset my password within the `dev` Firebase and got a 'insufficient permissions' error message in the console. [See video here](https://www.loom.com/share/3392949fc723444cb616248b62e6cd40) showing this. Any ideas why this would happen to a typical user @ibeckermayer ? (the user is my covidwatch email address). The error message points to the function `onSnapshot` defined in `model.js`.
> 
> Coming up to speed a bit here having not been too involved in email password set up workflow.

Yes, that's https://github.com/covid19risk/permission-portal/issues/227.

PR #293, opened 2020-06-23
Harmonize package.json license to Apache 2.0

aidan-fitz:
Partially a test to see if TSLint fails on all branches or just mine.


jcolla-holla:
This has been sitting here a while and makes sense to merge.  If ever we want to adjust our license we can do so easily.

PR #298, opened 2020-06-25
Create bespoke ChangeRoleModal

### LINKED ISSUE #273: Update the Manage Users "confirm change" from alert to bespoke Modal
Currently we ask users if they want to change other user's role's via the browser alert api. This should be converted to the bespoke Modal in Figma v3 (see under `Admin / Team Management` frame in the `Main Admin Screens` row).

aidan-fitz:
Closes #273

`ChangeRoleModal` looks like:
you can tell I've watched too much Avatar recently

To test: toggle a user's role between Org Admin and Validator to bring up the prompt, then click Confirm to ensure the role is changed or Cancel to exit.

### DIFF
--- a/frontend/client/Styles/application.scss
+++ b/frontend/client/Styles/application.scss
@@ -3,6 +3,7 @@
 @import 'screens/add_member_modal';
 @import 'screens/forgot_password_modal';
 @import 'screens/change_password_modal';
+@import 'screens/change_role_modal';
 @import 'components/footer';
 @import 'components/navbar';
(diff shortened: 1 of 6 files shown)


jcolla-holla:
I made a minor styling change as the text visually looked a bit cramped.  I'm not a designer but space it out a bit more and also bolded the user's name.  Deliberated if it made sense to use a period or a colon after the user's name, landed on period making the most sense!

Screen Shot 2020-06-25 at 9 27 51 PM

PR #307, opened 2020-07-01
Verification Server integration with cookie/session

colbymorrison:
The so-called cookie-server approach to integrating the verification server (VS) with the permission portal.

The idea is to have one user in the verification server for the whole permission portal. When a code is requested by a user from the frontend, the new cloud function `getVerficationCode` logs into the server as that user and gets a new code via http requests.

#### Requires 

- The verification server running locally (instructions are [here](https://github.com/google/exposure-notifications-verification-server), using either CW firebase project or your own)
- A local firebase project for the permission portal that is set up correctly (so we don't mess up the dev project)

#### Setup the local verification server
Picking up the last steps from the README of the exposure-notifications Repo linked above, create a user with email `**email**` for the verification server and start the server

```
go run ./cmd/add-users --email **email** --name "First Last" --admin true
go rum ./cmd/server
```

Log into the server on `localhost:8080` which will prompt you to create a password, which we'll refer to as `**password**`

Since we're going to call the server from a Firebase cloud function, we can't just call into `localhost`, as the function will be running on one of Google's cloud machines. 

So, I used one of my fav programs, [ngrok](https://ngrok.com/), to tunnel localhost out to the internet.

With the server still running, in a new terminal run
```
ngrok http -subdomain=gvs 8080
```

Now whatever is being served to `http://localhost:8080` on my machine, currently the verification server, is tunneled to `http://gvs.ngrok.io` on the outer internet! So I can go to `http://gvs.ngrok.io` and access my local verification server.

### Deploy the cloud functions

The cloud function gets the configuration it needs to login from firebase environment variables.
We control these in `config/functions.config.*.js`.

Since we're running this all locally for now edit the `verif_server` object in `config/functions.config.local.js`:
 - url: the ngrok url from above (for me `http://gvs.ngrok.io`), in prod this will be `verification.covidwatch.org`
 - email: `**email**` from above

The password and key fields are secret so we'll set them as environment variables (see commands below). In prod these will be supplied by GitHub secrets and the below commands are executed by GitHub actions on each deployment.  The key field is the "api key" for the Firebase Project running the Verification Server. It is whatever you set as the `FIREBASE_API_KEY` env var when setting up the server locally, call this `**key**`.

Copy the new config over and deploy the new env vars to our local firebase project:
```
cd cloud-functions/functions
npm run init
export VERIF_SERVER_PASSWORD=**password**
export VERIF_SERVER_FIREBASE_API_KEY=**key**
export NODE_ENV=local
npm run config:set
```

Deploy the new cloud functions

```
cd cloud-functions/functions
export NODE_ENV=local
npm run build
firebase deploy --only functions --project local
```

### Run the frontend app

Since we're already using port 8080 for the VS, we need to run the frontend on a new port.

Edit the port field `frontend/webpack.config.dev.js` to something new (ex. 1227)

With the VS and ngrok still running, run the frontend app for the local firebase project
```
cd frontend
npm run start:local
```

good

### DIFF
--- a/cloud-functions/functions/src/index.ts
+++ b/cloud-functions/functions/src/index.ts
@@ -4,8 +4,8 @@ import * as sgMail from '@sendgrid/mail';
 import { randomBytes } from 'crypto';
 import axios from 'axios';
 import axiosCookieJarSupport from 'axios-cookiejar-support';
-import toughCookie from 'tough-cookie';
-import queryString from 'query-string';
+import * as toughCookie from 'tough-cookie';
+import * as queryString from 'query-string';
 
 // Initialize firebse admin and get db instance
 admin.initializeApp(functions.config().firebase);
--- a/cloud-functions/functions/src/index.ts
+++ b/cloud-functions/functions/src/index.ts
@@ -2,6 +2,10 @@ import * as admin from 'firebase-admin';
 import * as functions from 'firebase-functions';
 import * as sgMail from '@sendgrid/mail';
 import { randomBytes } from 'crypto';
+import axios from 'axios';
+import axiosCookieJarSupport from 'axios-cookiejar-support';
+import toughCookie from 'tough-cookie';
+import queryString from 'query-string';
 
 // Initialize firebse admin and get db instance
 admin.initializeApp(functions.config().firebase);
@@ -442,17 +446,13 @@ export const initiatePasswordRecovery = functions.https.onCall((body) => {
 });
 
 export const getVerificationCode = functions.https.onCall(async () => {
-  const axios = require('axios');
-  const axiosCookieJarSupport = require('axios-cookiejar-support').default;
-  const tough = require('tough-cookie');
-  const qs = require('querystring');
   axiosCookieJarSupport(axios);
 
-  let config = functions.config().verif_server;
-  let url = config.url;
+  const config = functions.config().verif_server;
+  const url = config.url.slice(-1) === '/' ? config.url : config.url + '/';
 
-  const cookieJar = new tough.CookieJar();
-  let instance = await axios.create({
+  const cookieJar = new toughCookie.CookieJar();
+  const instance = await axios.create({
     jar: cookieJar,
     withCredentials: true,
   });
@@ -463,12 +463,12 @@ export const getVerificationCode = functions.https.onCall(async () => {
       { email: config.email, password: config.password, returnSecureToken: true }
     );
 
-    let form = { idToken: response.data.idToken };
+    const form = { idToken: response.data.idToken };
 
     // Get CSRF token
     response = await instance.get(url);
 
-    response = await instance.post(url + 'session', qs.stringify(form), {
+    response = await instance.post(url + 'session', queryString.stringify(form), {
       headers: {
         'Content-Type': 'application/x-www-form-urlencoded',
         'X-CSRF-Token': response.headers['x-csrf-token'],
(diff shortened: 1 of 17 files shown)


jcolla-holla:
Nice.  One thing to be aware of (though I don't think it matters for us) is that we are NOT reporting a test date w this approach.  The Google UI has this as optional (and hopefully doesn't become mandatory for some reason in the future).  I don't see any issue with us not reporting this or touching it at all, but just wanted to call attention to it.

config/functions.config.local.js
```diff
@@ -4,6 +4,12 @@ var config = {
     },
     "sendgrid": {
         "key": process.env.SENDGRID_API_KEY
+    },
+    "verif_server": {
+        "url": "**url**",
+        "email": "**email**",
+        "password": process.env.VERIF_SERVER_PASSWORD,
+        "key": process.env.VERIF_SEVER_FIREBASE_API_KEY
```
ibeckermayer, line 12: This isn't a secret key, so we can just include it directly in the config rather than make it an environment var.

cloud-functions/functions/src/index.ts
```diff
@@ -440,3 +440,50 @@ export const initiatePasswordRecovery = functions.https.onCall((body) => {
       });
```
ibeckermayer, line 440: Please apply `eslint` (`frontend/.eslintrc.js`) and `prettier` (`frontend/.prettierrc`) to this file, there are a bunch of formatting discrepancies, several let's that should be const's, etc.

cloud-functions/functions/src/index.ts
```diff
@@ -440,3 +440,50 @@ export const initiatePasswordRecovery = functions.https.onCall((body) => {
...
   });
 });
+
+export const getVerificationCode = functions.https.onCall(async () => {
+  const axios = require('axios');
```
ibeckermayer, line 445: Let's make these `ES6` style imports to stay consistent with the rest of the file.

cloud-functions/functions/src/index.ts
```diff
@@ -440,3 +440,50 @@ export const initiatePasswordRecovery = functions.https.onCall((body) => {
...
+
+  const cookieJar = new tough.CookieJar();
+  let instance = await axios.create({
+    jar: cookieJar,
+    withCredentials: true,
+  });
+
+  try {
+    let response = await instance.post(
+      'https://www.googleapis.com/identitytoolkit/v3/relyingparty/verifyPassword?key=' + config.key,
```
ibeckermayer, line 462: Out of curiosity, where did you find this url? Do others do this non-firebase-client firebase auth login elsewhere (I imagine so)?

cloud-functions/functions/src/index.ts
```diff
@@ -440,3 +440,50 @@ export const initiatePasswordRecovery = functions.https.onCall((body) => {
...
+      { email: config.email, password: config.password, returnSecureToken: true }
+    );
+
+    let form = { idToken: response.data.idToken };
+
+    // Get CSRF token
+    response = await instance.get(url);
+
+    response = await instance.post(url + 'session', qs.stringify(form), {
```
ibeckermayer, line 471: Can we optionally append a '/' character here before 'session' if it dne, since we both ran into that issue?

cloud-functions/functions/src/index.ts
colbymorrison: I got this from logging in to the Google Verif Server UI and looking at the network requests it made in the browser dev tools.

ibeckermayer: Nice

colbymorrison: FWIW [here](https://developers.google.com/identity/toolkit/web/reference/relyingparty/verifyPassword) is the official documentation for this API endpoint

PR #308, opened 2020-07-02
Sanity check on signup protection

ibeckermayer:
The test to look at is called `'An attacker cannot create an account and then quickly create their own record in /users table before onCreate has time to run and delete them'`, the rest of the changes are formatting differences (argh!).

I thought up a potential attack where a user creates an account (in Firebase auth) through the client library, then quickly signs in and creates their own user document before our `onCreate` hook has time to run and delete their account. As it turns out, our `isAdmin()` rule protects against this, but I added a test to assure this is enforced.

### DIFF
--- a/cloud-functions/functions/__tests__/auth.test.ts
+++ b/cloud-functions/functions/__tests__/auth.test.ts
@@ -53,7 +53,7 @@ afterEach(async () => {
   }
 });
 
-test('clientAuth.createUser winds up with a deleted user even if they try to quickly create their own record in /users table', async () => {
+test('An attacker cannot create an account and then quickly create their own record in /users table before onCreate has time to run and delete them', async () => {
   // Attacker creates an account with the client auth
   await clientAuth.createUserWithEmailAndPassword('shouldBe@deleted.com', 'shouldBe@deleted.com');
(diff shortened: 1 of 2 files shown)


cloud-functions/functions/__tests__/auth.test.ts
```diff
@@ -56,6 +53,28 @@ afterEach(async () => {
   }
 });
 
+test('An attacker cannot create an account and then quickly create their own record in /users table before onCreate has time to run and delete them', async () => {
```
aidan-fitz, line 56: "before onCreate has time to run and delete them" is confusing. Why does the record have to be deleted before we can create it?

ibeckermayer: `onCreate` is a cloud function that's called whenever a user is created in Firebase Auth, that we're using in part to delete any users that register with Firebase Auth but aren't in our `/users` collection. That way, public sign ups are effectively "turned off", even though Firebase Auth doesn't support that as a feature.

This is a simple attack I thought of to try and circumvent that system. Maybe the test could be worded better, though I think in context of understanding our system it makes sense (?)

PR #310, opened 2020-07-06
Add email to sign in link url

ibeckermayer:
Both password reseters and new users can now click the link in their email and be immediately signed in, and asked to change their passwords:

Screen Shot 2020-07-06 at 11 53 02 AM

closes https://github.com/covidwatchorg/permission-portal/issues/304

### DIFF
--- a/frontend/client/src/screens/Login.js
+++ b/frontend/client/src/screens/Login.js
@@ -58,9 +58,9 @@ const SignInFormBase = observer(
       // Based on https://firebase.google.com/docs/auth/web/email-link-auth
       if (this.props.store.isSignInWithEmailLink(window.location.href)) {
         Logging.log('signInWithEmailLink detected')
+        var email = new URL(location.href).searchParams.get('email')
         // Get the email if available. This should be available if the user completes
         // the flow on the same device where they started it.
-        var email = window.localStorage.getItem('emailForSignIn')
         if (!email) {
           // User opened the link on a different device. To prevent session fixation
           // attacks, ask the user to provide the associated email again. For example:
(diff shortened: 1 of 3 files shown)


frontend/client/src/screens/Login.js
```diff
@@ -58,9 +58,9 @@ const SignInFormBase = observer(
       // Based on https://firebase.google.com/docs/auth/web/email-link-auth
       if (this.props.store.isSignInWithEmailLink(window.location.href)) {
         Logging.log('signInWithEmailLink detected')
+        var email = new URL(location.href).searchParams.get('email')
         // Get the email if available. This should be available if the user completes
         // the flow on the same device where they started it.
-        var email = window.localStorage.getItem('emailForSignIn')
         if (!email) {
           // User opened the link on a different device. To prevent session fixation
           // attacks, ask the user to provide the associated email again. For example:
```
aidan-fitz, line 66: This doesn't work anymore. Now that we are reading the email from the URL instead of local storage, we cannot guarantee that the user is logging in from the device that generated the magic link.

frontend/client/src/screens/Login.js
```diff
@@ -58,9 +58,9 @@ const SignInFormBase = observer(
       // Based on https://firebase.google.com/docs/auth/web/email-link-auth
       if (this.props.store.isSignInWithEmailLink(window.location.href)) {
         Logging.log('signInWithEmailLink detected')
+        var email = new URL(location.href).searchParams.get('email')
         // Get the email if available. This should be available if the user completes
         // the flow on the same device where they started it.
-        var email = window.localStorage.getItem('emailForSignIn')
         if (!email) {
```
aidan-fitz, line 64: This is dead code. Magic links will always have an `email` query string parameter.

aidan-fitz:
This violates Firebase's guidelines for authentication:

> Do not pass the user’s email in the redirect URL parameters and re-use it as this may enable session injections. ([magic link security concerns])

We can and should still refactor the magic link logic into `generateSignInWithEmailLink` as you did in this PR, so we can use it with both initial sign in and password reset.

[magic link security concerns]: https://firebase.google.com/docs/auth/web/email-link-auth#security_concerns

PR #311, opened 2020-07-08
Error handling for magic links

### LINKED ISSUE #309: Alert user if signInWithEmailLink fails
In `Login.js`, if the call to `signInWithEmailLink` fails we should show the user a toast telling them "The link you clicked has expired. Please sign in with your password, or restart the password recovery process."

aidan-fitz:
Closes #309.


aidan-fitz:
Newest commit: refactor like in #310

PR #327, opened 2020-07-14
V4 login

jcolla-holla:
A question I have: are we effectively hard coding university imagery for v4?  Like this image of students gathered together.  Seems the answer is yes for short term.

Adjusted `App.js` routing a bit in order to render a consistent NavBar across all pages as now the NavBar is identical on all pages.

The Login page consistently on multiple screens I tested leaves a tiny bit of space for the user to scroll up and down, which leaves a bit more space under the footer.  However, we could easily change this so that the user has no where to scroll up or down and bring all sections a bit closer together vertically.

## Before

Screen Shot 2020-07-14 at 1 33 03 PM

## After 

Screen Shot 2020-07-14 at 3 54 51 PM


jcolla-holla:
> Looks good, yes we're just hardcoding the uofa branding for v4. Let's just get rid of that extra scroll-down like you mentioned before merging.

Ok, updated that.  See if this has the effect you have in mind locally.  There is not a ton of space underneath the footer text now, which maybe is the desired output.  We can play with this as needed.

PR #332, opened 2020-07-15
v4 Manage Members page and other stuff

ibeckermayer:
- Changed all `#333333` and` #585858` to `#383838 `
- removed box shadows
- engaged in all sorts of grotesque css wrangling to get things to look correct

Our css should definitely make a lot more sense than it currently does. This is not the PR that does that.

The way I QC'd this one was to walk through all the different pages / modals / etc. and ensure that they're in-line with the broad styling updates for v4, namely (quoting Vic) "titles, buttons, input fields, etc. shorter/slimmer... removed interior shadows..."

closes https://github.com/covidwatchorg/permission-portal/issues/320


ibeckermayer:
> I understand that we are opting for an 'Add Member' button and modal.

Correct

> * Need more vertical space between the "Manage Members" `<h1>`, "Add Member" button, and grid

- [ ] TODO

> * Border radius of the grid should be less to match design

- [ ] TODO

> * The pagination numbers at bottom right of grid don't match the design

- [ ] TODO

> * I don't see a search bar

This is a major technical project, just commented in the Figma that we should hold off on this for now

> * Success/Failure message doesn't appear directly under NavBar now (this is probably true across all pages since the NavBar height changed)

https://github.com/covidwatchorg/permission-portal/issues/334

> * The Add Member modal doesn't quite follow the v4 design (main header size, the padding for the X, etc. -- feel free to create another issue for refactoring Modals across the board if that makes sense)

- [ ] TODO

PR #338, opened 2020-07-16
changing pagination arrows and css to match figma

ibeckermayer:
closes https://github.com/covidwatchorg/permission-portal/issues/335


jcolla-holla:
The gray arrow piece now looks like this in this PR:

Screen Shot 2020-07-17 at 1 46 04 PM

ibeckermayer:
> ### EDIT: I just completed this and pushed it to this branch and approved it. See next comment
> The arrows of this PR are misaligned with the v4 design and what is currently in staging (what is in staging appears to be the same as the v4 design). So in order for us to merge this PR, I suggest we bring in the code from what is currently on branch dev for the custom arrow dropdown for elements. To avoid overwriting that with something worse.
> 
> FYI I'm doing this now so no need to work on it, I'm just calling it out. See next comment
> 
> Visual here:
> 
> ## Staging
> Screen Shot 2020-07-17 at 1 14 14 PM
> 
> ## This PR
> Screen Shot 2020-07-17 at 1 14 10 PM

This PR is for the pagination arrows at the bottom right, it shouldn't affect those arrows.

jcolla-holla:
Issues I see to resolve:

- [ ] Visually there appears to be more vertical AND horizontal padding for each cell in this PR than in the design.  I've noticed that sometimes the Figma CSS seems to be off in terms of what we want to achieve so in this case if that's what was used we may need to decrease the padding/margin values slightly to achieve the visual result showed in Figma.

- [ ] The column headers are sometimes misaligned with the cells. (image below)

Screen Shot 2020-07-17 at 2 08 17 PM

PR #345, opened 2020-07-17
V4 code validations dates

jcolla-holla:
Some minor workflow updates to the `CodeValidations.js` to maintain dates between today and 14 days ago.

Updates:

1) show Toast msg on choosing date out of bounds
2) do not allow code generation until a valid date is chosen
3) return to the code format 000-000-00 instead of 0000-0000

Preview:
ezgif com-video-to-gif

### DIFF
--- a/frontend/client/src/screens/CodeValidations.js
+++ b/frontend/client/src/screens/CodeValidations.js
@@ -50,15 +50,14 @@ const CodeValidationsBase = observer((props) => {
   const handleDate = (e) => {
     e.target.classList.add('with-value')
     e.target.classList.remove('no-value')
-    const todayDate = new Date();
 
     // this logic here needs to be cleaned up.  specifically the moreThanFourteenDaysAgo and dateInFuture functions in time.js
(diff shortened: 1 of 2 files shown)


frontend/client/src/screens/CodeValidations.js
```diff
@@ -26,6 +28,19 @@ const CodeValidationsBase = observer((props) => {
 
   let confirmedToast = createRef()
 
+  useEffect(() => {
+    updateButtonDisabled()
+  })
+
+  // Updates the buttonDisabled variable state based on other state variables
+  const updateButtonDisabled = () => {
+    if (needsReset || testType === '' || dateInvalid) {
```
jcolla-holla, line 37: I believe one final option is needed in this conditional in order to prevent a situation where the generate code button is enabled by clicking a diagnosis radio but NOT filling out a date:

`|| (testType !== '' && testDate === '')`

So, the if statement on line 37 would read:

`if (needsReset || testType === '' || dateInvalid || (testType !== '' && testDate === ''))`

I tested this locally and I believe it protects our state the way we intend.

PR #353, opened 2020-07-18
Countdown - nice to have - anyone feel free to contribute

jcolla-holla:
I was playing around with having a live countdown without refreshing of the time left for a generated code to be valid on `CodeValidation.js`.  This is just a sandbox for a non-urgent feature, if someone else wants to take this across the finish line go for it.


jcolla-holla:
Gonna close this actually for the time being as it may be a distraction from more urgent needs

PR #369, opened 2020-07-21
update getVerificationCode

ibeckermayer:
Modifies the frontend and backend for the latest version of the VS, which allows us to use an api key to hit an admin server to issue tokens.

Also updates `testDate` to be `symptomDate` per the updated VS api.

### DIFF
--- a/config/functions.config.local.js
+++ b/config/functions.config.local.js
@@ -5,11 +5,9 @@ var config = {
   sendgrid: {
     key: process.env.SENDGRID_API_KEY,
   },
-  verif_server: {
-    url: "https://verification.covidwatch.org/",
-    email: "john@covidwatch.org",
-    key: "AIzaSyCQFrBAHQ1AJ5ayDYah7LPF-OgoUxqFZTs",
-    password: process.env.VERIF_SERVER_PASSWORD,
+  verification_server: {
+    url: process.env.VERIFICATION_SERVER_URL,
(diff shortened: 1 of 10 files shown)


config/functions.config.staging.js
```diff
@@ -1,15 +1,13 @@
 var config = {
   client: {
-    url: "http://localhost:3000/",
+    url: "https://staging-permissions.covidwatch.org/",
```
colbymorrison, line 3: Should we also update this field in the prod file?

colbymorrison:
Looks good so long as that axios request works correctly. I don't have the new verif server set up so I didn't test locally.

config/functions.config.staging.js
jcolla-holla: Given this is the file `config/functions.config.staging.js` shouldn't this be our staging URL? permission-portal.covidwatch.org.  

Accordingly, shouldn't `config/functions.config.prod.js` have the prod URL: portal.covidwatch.org?

config/functions.config.local.js
```diff
@@ -5,11 +5,9 @@ var config = {
   sendgrid: {
```
jcolla-holla, line 5: right above here should update the URL to portal.covidwatch.org, right?  Or do we want to save that update for when we are ready to actually push this to prod and keep the scope of this PR more staging?  Not sure if matters either way.

Issue #373, opened 2020-07-22
Swap out inline styles that use any colors in color.scss


ibeckermayer:
Move these styles to scss files so that they can be centrally managed

Issue #372, opened 2020-07-22
remove all instances of useStyles in exchange for scss


ibeckermayer:
For the sake of consistency

colbymorrison:
+1 for this

colbymorrison:
Did this actually get resolved? I think it may have been closed on accident.

PR #356, opened 2020-07-20
Refactor CSS for Text Styles

colbymorrison:
Closes #318 and Closes #352. (This probly should be 2 PRs... whoops)

- Sets common text styles in `application.scss` based on the styles from the v4 figma. These styles are the various header sizes, body text, small text, and extra small text. Refactors other files to use these classes whenever possible, only overriding when necessary. Eliminates a lot of unnecessary lines like `font-family`, `font-style`, and `font-size` in the css files. 

- **UPDATE** `application.scss` now contains default styles for `anchor`, `label`, `select` and `input` tags based on v4 figma. Again, updates styles to only override these when necessary. 

- Edits styling on the Manage Settings page to better match Figma. 

- Adds consistent margin between NavBar and top of each page. 

- Updates menu styling. 

- Ensures pages don't extend beyond hamburger menu.

- Fixes a small problem I saw on the Login page where the grey box was really stretched out on larger screens:
### Before
login-old
### After
login-new

Also snuck in a couple of updates to the README.

PR #374, opened 2020-07-22
Css color refactor

ibeckermayer:
Uses Sass variables for the main colors rather than copying them throughout the various sass files. There are still inline styles which use the hardcoded `#*`, see (https://github.com/covidwatchorg/portal/issues/372 and https://github.com/covidwatchorg/portal/issues/373)

There was some `webpack` weirdness going on that I wasn't able to resolve, see the `// NOTE:`'s I left in the comments.

This is branched off of @colbymorrison's [PR](https://github.com/covidwatchorg/portal/pull/356) and so we should wait until that one's merged before merging this one.


frontend/client/src/components/ChangeRoleModal.js
```diff
@@ -16,7 +16,7 @@ const ChangeRoleModal = withStore((props) => {
 
   return (
     <Modal hidden={props.hidden} onClose={props.onClose} containerClass="change-role-modal-container">
-      <h2>Confirm Change</h2>
+      <h3>Confirm Change</h3>
```
jcolla-holla, line 19: If we are using `<h3>` for modal titles, let's use that consistently across all modals.  So for example line 106 of `ChangePasswordModal.js` I believe is using an `<h2>`, when it should be an `<h3>`

colbymorrison:
Looks good!

frontend/client/src/components/ChangeRoleModal.js
colbymorrison: I agree with this, all modal headers should be `<h3>`

PR #380, opened 2020-07-22
add footer links (Privacy Policy, TOS)

### LINKED ISSUE #367: Get footer links and make them all pop open in new tabs
We need to decide on where we're going to send users for `Terms of Use` and `Support`.

All of three of these should pop open in new tabs (currently `Privacy Policy` doesn't)

jcolla-holla:
Closes #367 .  Still waiting on some clarification on the Terms of Service document so this will be a draft for now until that is finalized.  Then should be quick merge given minimal adjustments happening here

### DIFF
--- a/frontend/client/src/components/Footer.js
+++ b/frontend/client/src/components/Footer.js
@@ -11,10 +11,10 @@ const Footer = () => {
         </div>
         <div id="footer-text">
           <p id="copyright">© 2020 Covid Watch. All rights reserved.</p>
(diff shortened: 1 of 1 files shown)


jcolla-holla:
I removed the Terms of Use link because it seems that it will take some time to receive this from legal team and currently we have a dead link for it in production.  Obviously having dead links is bad so better to remove it for now to avoid confusing user and we can update when we have a Terms of Use document in hand.

PR #381, opened 2020-07-22
Manage Members/Add Members down arrow fix

### LINKED ISSUE #376: make down arrows click-able on Manage Members and Add Member Modal
Currently they don't allow you to click them

leon-i:
Resolves   #376 
  
Improved padding on select elements to make them fill their container better, made down arrows clickable.  
  
Image preview of the change:  
image

### DIFF
--- a/frontend/client/Styles/screens/manage_teams.scss
+++ b/frontend/client/Styles/screens/manage_teams.scss
@@ -45,7 +45,7 @@ table {
         background-color: #ffffff;
         border-radius: 0;
         select {
-          padding: 0 10px;
+          padding: 5px 10px;
           border: none;
           width: 100%;
           height: 100%;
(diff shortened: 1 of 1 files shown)


frontend/client/Styles/screens/manage_teams.scss
```diff
@@ -70,9 +70,10 @@ table {
         content: url('../../assets/arrow-down.svg');
         position: absolute;
         right: 10px;
-        top: 7px;
+        top: 10px;
         border-radius: 0;
         background-color: transparent;
+        pointer-events: none;
```
jcolla-holla, line 76: @leon-i What is the logic for not having any pointer events?

leon-i: From my understand the `:after` that constitutes the down arrow was not clickable because it resided on top of the select and had no mouse events associated with it. Granting it `pointer-events: none` allows it to be clicked through, giving the desired effect.

PR #386, opened 2020-07-22
add consistent success toast language of 'Success: xyz'

### LINKED ISSUE #385: Toast text should be smaller, aligned w v4 Figma

jcolla-holla:
Closes #385 

I tried to make the messaging also a bit more consistent in its format.  So for any successful toast messages, it begins with "Success: ...".  And the only letter capitalized is the first letter of each sentence.

The images may not showcase this obviously but the text is smaller and less intimidating.

## Before:
Screen Shot 2020-07-22 at 4 09 04 PM

## After:

Screen Shot 2020-07-22 at 4 09 44 PM


frontend/client/src/screens/ManageTeams.js
```diff
@@ -33,15 +33,15 @@ const ManageTeamsBase = observer((props) => {
...
+    setToastMessage('Success: member email invitation sent')
     setIsSuccess(true)
     setShowAddMemberModal(false)
     confirmationToast.current.show()
   }
 
   const onAddMemberFailure = (e) => {
     Logging.error(e)
-    setToastMessage('Member Email Invitation failed to send')
+    setToastMessage('Member email invitation failed to send')
```
colbymorrison, line 44: For this one, should we give a more descriptive message based on the error? For example, this can occur if you try and add a user that already exists. Perhaps in that case the Toast could say "Member email invitation failed to send, user already exists." or something like that.

jcolla-holla: Agreed.  See my [most recent commit](https://github.com/covidwatchorg/portal/pull/386/commits/da9db9f8c2317c6b09496cf85ddbde95b0796d88) which addresses this case.  If that is the only concern, feel free to approve and merge!

PR #389, opened 2020-07-23
Remove unused package file-system

aidan-fitz:
`file-system` contains [a vulnerability](https://www.npmjs.com/advisories/1502) that hasn't been addressed in over 4 months. Since we're not using it anyway, this vulnerability doesn't affect us, but at least it won't bother `npm audit`.


ibeckermayer:
How do you know we're not using it?

Issue #349, opened 2020-07-17
AddMember modal should have the same <select> custom arrow as table


jcolla-holla:
This arrow doesn't match the custom gray arrow used in the Manage Members table

<img width="418" alt="Screen Shot 2020-07-17 at 2 11 21 PM" src="https://user-images.githubusercontent.com/16062590/87831093-8a36af80-c850-11ea-9825-ccadae1c98b0.png">

PR #392, opened 2020-07-23
Add member select arrow

ibeckermayer:
closes https://github.com/covidwatchorg/portal/issues/349


frontend/client/src/components/AddMemberModal.js
```diff
@@ -147,7 +148,9 @@ const AddMemberModalBase = observer((props) => {
         <label htmlFor="role" style={{ marginTop: state.emailValidationFailed ? '6px' : null }}>
           Role<span>*</span>
         </label>
+        {/* <div className="custom-select"> */}
```
jcolla-holla, line 151: these commented out divs should not be commented out, right?

frontend/client/styles/screens/add_member_modal.scss
```diff
@@ -56,6 +57,13 @@
     font-size: 16px;
     padding-left: 20px;
   }
+  .custom-select:after {
+    content: url('../../assets/arrow-down.svg');
+    position: relative;
+    right: 25px;
+    top: -1px;
+    background-color: transparent;
```
jcolla-holla, line 65: The arrow itself is currently not clickable, which could be frustrating for the user.  Take a look at Isak's PR [here](https://github.com/covidwatchorg/portal/pull/381) for how this was solved on the Manage Team page.  I think it may just be the `pointer-events: none` property to the `:after`

ibeckermayer:
> A few things to change, see my comments in various parts of the code.
> 
> And even though the goal of this PR is not to solve this type of thing necessarily, maybe include some padding-left on the main header of the modal as it is poorly placed currently (see image below)
> 
> Screen Shot 2020-07-23 at 4 42 46 PM

Created an issue for this and so am just going to put it off for another PR. The modal css general needs a re-think

Issue #377, opened 2020-07-22
Only ask for the email in an alert if not firstTimeUser


ibeckermayer:
@ibeckermayer knows what that means

ibeckermayer:
Modification: just send new users to the website and require that they log in with their temporary password.

PR #396, opened 2020-07-23
remove auto sign in link from new user email

ibeckermayer:
Closes https://github.com/covidwatchorg/portal/issues/377 (see comment)


ibeckermayer:
> I don't see anything of issue here.
> 
> Without testing it, I'm assuming the `functions.config().client.url` on line 183 returns the staging or prod URLs, right?

Yep

PR #408, opened 2020-07-25
Refactor Modals

### LINKED ISSUE #300: Better "Componentize" Modals
Modals have common functions -- Title, subtext, a stack of inputs, some inputs have rules, submit buttons, optional dismissibility, etc.

Let's make them compose-able from props to make them easier to build and maintain.

### LINKED ISSUE #306: Refactor CSS for modals
All modals should have a padding and consistent styles for headings and paragraphs so that we don't have to add weird margins for elements inside the modals.

Follow-up to #305

colbymorrison:
Closes #300 
Closes #306 

"Componentizes" modals and refactors their CSS

- All modals have a title, so I added it as a prop to `Modal.js`
- `modal.scss` now contains styling for modal titles and PendingOperationButtons, so their margins are consistent across modals. It also now has styling for modal width and padding, so those are also consistent. 
- Creates a new component called `ModalInput` to encapsulate a `label` and `input` in a Modal, such as this:
modal_input
Its props include: if the input is required, a function to check the input is valid, and a message to display under the text box if the input is invalid. See the comment in `ModalInput.js` for a description of all props. `modal_input.scss` allows us to change the styling for `ModalInput`s in one place. This really cleans up the modal css iles.
- Creates a new component `CancelConfirmButtons` for use in `ChangeRoleModal` and `ChangeStatusModal` (if we implement the updated add user flow in the Figma, this can be used there too)


colbymorrison:
Updated:

1) Added a div as you suggested to not push elements down on error text. Example below for Forgot Password and Add Member modals:
Screen Capture_Navigator_20200727213734

2) Reset image modal:
reset

3) Change password modal:
reset-pass

jcolla-holla:
Nice.  I see a few more things that need addressing:

- [ ] The "Uploading..." text and loading graphic is not centered on uploading a picture.  And the original modal header remains.  I think it's ok if the original header remains but it may be better to remove it in favor of only "Uploading...".

Screen Shot 2020-07-27 at 10 26 01 PM

- [ ] The `font-weight` of modal buttons (at least a few I looked at) is `bold` but in other cases buttons have a value of `600`.  `bold` is a value of `700` so there is a slight inconsistency between modals and other buttons in the app.  Let's have all buttons have a `font-weight` of `600` (I believe this is in line w Figma)

colbymorrison:
Updated "Uploading" Screen:
uploading

Changed default font weight for buttons to 600 in `modal.scss`

colbymorrison:
Updated the button style on that Modal to `save-button`, the same as the other modals
pass

frontend/client/src/components/Modal.js
```diff
@@ -19,6 +24,7 @@ const Modal = (props) => {
             <img src={close} alt="Close" />
           </span>
         )}
+        <h3>{props.title}</h3>
         {props.children}
```
ibeckermayer, line 28: I think we can remove this, doesn't seem to be used anywhere else

frontend/client/src/components/ModalInput.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react'
+
+/**
```
ibeckermayer, line 3: Good documentation

jcolla-holla:
> Great job, ready to merge assuming all of @jcolla-holla's critiques are accounted for. Just requested to remove an unused line (see my inline comments)

Yup, all addressed.  Looks like the final touches are what your comments point out.

frontend/client/src/components/Modal.js
ibeckermayer: I was wrong, it turns out `props.children` is a special value, see [here](https://reactgo.com/react-props-children/)

PR #414, opened 2020-07-28
remove permissions urls in functions for sendgrid email

jcolla-holla:
As the permissions URLs are no longer available via AWS.  I maybe should have set up forwarding but in lieu of that modifying the sendgrid URL here.


jcolla-holla:
I'm merging this without review to test in staging

PR #418, opened 2020-07-28
Closing Add Member Modal Now Resets State

### LINKED ISSUE #417: Add Member Modal data should be cleared when exited
When I add multiple members to the portal, it doesn't clear the info from the first member I added in the "Add member" modal

leon-i:
Resolves #417  
  
Adding a member now clears the state so their information does not appear again when opening modal. Closing a modal while in progress of adding a member now resets state as well.

### DIFF
--- a/frontend/client/src/components/AddMemberModal.js
+++ b/frontend/client/src/components/AddMemberModal.js
@@ -34,21 +34,23 @@ const ValidationRules = [
   },
 ]
 
+const defaultState = {
+  firstName: '',
+  firstNameValidationFailed: false,
+  firstNameValidationMessage: '',
+  lastName: '',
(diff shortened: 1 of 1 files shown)


leon-i:
Weird - I wasn't getting that locally. I'll look into it.

leon-i:
This is how it looks for me locally:  
  
image  
  
Perhaps the issue is browser based? I'm using chrome.

colbymorrison:
Hm yeah, since it looks that way for you I’d guess it is a browser thing or something related to that, I’m using Firefox.

ibeckermayer:
Added an issue [here](https://github.com/covidwatchorg/portal/issues/431)

PR #419, opened 2020-07-28
Reset Password Fix

### LINKED ISSUE #416: When resetting password, new password can't be same as curreent password
From Jeff

leon-i:
Resolves #416  
  
Reset password modal now rejects your new password if it is the same as the inputted current password. Updated `onChange` function of `ResetPasswordModalBase` to use a switch statement for better readability/efficiency.  
  
Note: this does not fix issues associated with the "forgot your password" reset via the login page (you are able to use the same password as the previous).

### DIFF
--- a/frontend/client/src/components/ResetPasswordModal.js
+++ b/frontend/client/src/components/ResetPasswordModal.js
@@ -35,21 +35,27 @@ class ResetPasswordModalBase extends React.Component {
     // Serialize updates
     this.setState((state) => {
       let newState = {}
-      if (fieldName === 'current-password') {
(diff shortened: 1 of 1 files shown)


colbymorrison:
Looking back at Slack, this issue was in regard to the new user case, when we ask new users to change their password on their first login. This uses the same code as the Forgot Password case, both use `ChangePasswordModal`. I think this PR is still good, but if we merge as is, we should make another issue for those cases.

ibeckermayer:
> ChangePasswordModal

See: https://github.com/covidwatchorg/portal/issues/416#issuecomment-665756640

PR #421, opened 2020-07-29
Ability to set custom logo

jcolla-holla:
First pass at setting a custom logo for teams via a script run by Covid Watch locally.  I got it working successfully with .png files (.svg and .jpeg tests were not successful yet and going to bed now).

The Login page now always has the Covid Watch logo in the top left and shows the user's organization logo upon logging in.  See gif example.

I left a lot of comments in the `custom-logo.js` file that is the script to run.  Test it out locally by switching between the .png files I linked for Soylent Green and Pepsi on your desktop (simulating the real situation of downloading some customer's logo and uploading that based on its file path on your desktop).

Feedback welcome.

ezgif com-video-to-gif

note: using Pepsi as a logo is just as an example


cloud-functions/functions/custom-logo.js
```diff
@@ -0,0 +1,50 @@
+// this is meant to be run by Covid Watch devs to set a custom logo for different organizations to appear in the NavBar
```
ibeckermayer, line 1: Let's call this something like `super-admin.js` so it makes sense to add other super admin functions to it as needed.

cloud-functions/functions/custom-logo.js
```diff
@@ -0,0 +1,50 @@
...
+// perhaps at some point this could be adapted for an admin user to set their organization's logo
+
+const admin = require('firebase-admin');
+const fs = require('fs');
+
+// these desktop paths for testing purposes, they work!  simulating a dev downloading a file from browser, saving to desktop, then referencing in this script to upload
+const soylentLogoPath = '../../../../soylent-green-logo.png'; // https://fanart.tv/fanart/movies/12101/hdmovielogo/soylent-green-51744f6296779.png
```
ibeckermayer, line 8: Let's not have these values hardcoded in the final commit

cloud-functions/functions/custom-logo.js
```diff
@@ -0,0 +1,50 @@
...
+
+// from some tests, svgs and jpegs don't work. line 39 hardcodes to png based base64 so that may be part of it
+const uOfAPath = '../../frontend/client/assets/uofa-logo.svg';
+const jpegExample = '../../../../Jesse_Colligan.jpeg';
+
+// note: environment should be manually set to "dev", "staging", or "production", depending on where you would like to update the logo
+const setLogo = (environment, organizationID, logoPath) => {
+  // Initialize admin SDK
+
+  // note: only the dev-firebase-admin keys are found in this directory.  staging and prod are elsewhere so this path is not correct for staging and prod
```
ibeckermayer, line 19: This isn't true, its up to the user to put the keys in this directory. If these dev keys are committed to the repo, we need to revoke them.

cloud-functions/functions/custom-logo.js
```diff
@@ -0,0 +1,50 @@
...
+  const serviceAccount = require(`../../permission-portal-cloud-functions/functions/permission-portal-${environment}-firebase-admin-key.json`);
```
ibeckermayer, line 20: Probably want to pull this stuff out of this function, so that we don't need to keep repeating it as we add other functions. dev/test/staging/prod could be set by a command line argument (preferable), or pulled from `process.env.NODE_ENV` (as it's done in reset-infrastructure.js)

jcolla-holla: Ok, a commit I just pushed allows for command line arguments.  See comments at the top of `super-admin.js` for example.

cloud-functions/functions/custom-logo.js
jcolla-holla: removed

cloud-functions/functions/custom-logo.js
jcolla-holla: Yup.  I believe I only have the dev json locally, I don't have the staging or prod files currently.  So will need the staging one in order to test further in staging before attempting to adjust in prod.

jcolla-holla:
### Note that if `organizationID` is invalid, Firebase supplies us with an error in the CLI:

Screen Shot 2020-07-29 at 1 45 08 PM

### Similar situation if `logoPath` is invalid, we get a helpful error in the CLI:

Screen Shot 2020-07-29 at 1 45 04 PM

I'm mentioning this because there is only one custom error message and that is for validating `environment`.  The other two arguments passed into the CLI for running `super-admin.js` are covered in terms of validation by pre-existing errors.

cloud-functions/functions/super-admin.js
```diff
@@ -0,0 +1,58 @@
...
+  fs.readFile(logoPath, (err, data) => {
+    if (err) {
+      console.log("setLogo only accepts .png currently, other file extensions will likely cause errors")
+      throw err;
+    }
+    const logo = data.toString('base64');
+
+    organization.update({
+      logoBlob: "data:image/png;base64," + logo,
+    });
```
ibeckermayer, line 47: Let's add something like
```javascript
.then(() => {
        console.log('Successfully updated logo');
      })
      .catch((err) => {
        console.error(err);
      });
```
here so we can get some immediate feedback

.gitignore
```diff
@@ -1,3 +1,7 @@
 .DS_Store
 **node_modules/**
-.vscode
\ No newline at end of file
+.vscode
+dist
+package-lock.json
+permission-portal-cloud-functions
```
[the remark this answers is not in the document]
jcolla-holla: I think I have some files in my repo from a way earlier version of the repo that have since been removed from `dev`.  And my git pulls don't update them so just making sure that they don't get committed by accident.  

I was thinking maybe I should just clone the `dev` repo again and start over instead of using this .gitignore stuff.

cloud-functions/functions/super-admin.js
```diff
@@ -0,0 +1,58 @@
...
+const setLogo = () => {
+  const environment = process.argv[2]
+  const organizationID = process.argv[3]
+  const logoPath = process.argv[4]
+
+  if (environment !== 'dev' && environment !== 'staging' && environment !== 'production') {
+    throw new Error("Invalid environment argument. Use 'dev', 'staging', or 'production' as the first argument in the format: node super-admin.js dev nN4h53UanIdvjMcf3gPm ../../../../pepsi-logo.png");
+  }
+
+  const serviceAccount = require(`../../permission-portal-cloud-functions/functions/permission-portal-${environment}-firebase-admin-key.json`);
```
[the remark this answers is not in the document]
jcolla-holla: The path confusion is probably related to my comment above about the .gitignore.

cloud-functions/functions/super-admin.js
jcolla-holla: Agreed, updated in new commit

PR #439, opened 2020-07-31
Minor adjustments to spacing for logo and footer links

jcolla-holla:
Based on recent changes to logo and footer links, needed to adjust spacing slightly of both to look natural.

1) Logo vertically centered more precisely
2) Footer links at bottom right justified right to be aligned with margin-right of each page

## Before

Screen Shot 2020-07-31 at 10 20 39 AM

## After

Screen Shot 2020-07-31 at 10 20 27 AM

### DIFF
--- a/frontend/client/App.js
+++ b/frontend/client/App.js
@@ -20,35 +20,31 @@ const App = () => {
         <Route path={ROUTES.NOT_FOUND}>
           <Fragment>
             <NotFound />
-            <Footer />
           </Fragment>
         </Route>
         <Route path={ROUTES.SETTINGS}>
           <Fragment>
             <Settings />
-            <Footer />
           </Fragment>
         </Route>
         <Route path={ROUTES.CODE_VALIDATIONS}>
           <Fragment>
             <CodeValidations />
-            <Footer />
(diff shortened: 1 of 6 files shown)


jcolla-holla:
I originally just added the `className` ternary to `Footer.js` based on what `window.location.pathname` is to adjust the class depending on if the user is logged in or out but that means that when logging out the class isn't adjusted because the Footer actually isn't re-rendered when each page loads (including logging out).  So using the `className` solution it was necessary to make sure the Footer renders upon logout.

It is probably possible (and perhaps cleaner) to use state for prompting re-renderings also as some other components use `props.store.data.user.isSignedIn` as the test for whether the user is signed in or not.  (like `ManageTeams.js`).  If that's cleaner to avoid the repetitive `<Footer />` in `app.js`, I can adjust in that direction though the current solution as is achieves the desired outcome I believe.

colbymorrison:
Did a quick experiment using `props.store.data.user.isSignedIn` instead of `window.location.pathname` to decide the class and it worked so I committed that change. I like this way a little more as it is more consistent with other pages.

PR #440, opened 2020-07-31
Removed Instances of useStyles, Replaced With SCSS

### LINKED ISSUE #420: Remove all instances of useStyles in exchange for scss
Duplicate of #372 which (I think) accidentally got closed as I still see `useStyles` throughout the codebase

leon-i:
Resolves #420  
  
Removed all instances of `useStyles` I could find, including in the `sample.js` component (thought I'm not sure where that appears in production).


frontend/client/src/screens/Settings.js
```diff
@@ -241,7 +215,6 @@ const SettingsBase = observer((props) => {
                 required
                 disabled={true}
                 aria-required="true"
-                className={input.root}
                 style={{ backgroundColor: '#e0e0e0' }}
```
colbymorrison, line 218: There are still quite a few lines like this around (style=[something]) that I think can also be moved to scss files. Especially ones like line 158 of this file where lots of styling is done inline.

colbymorrison:
Nice! I pushed a commit to remove `sample.js` since I agree it isn't used anywhere. @ibeckermayer just want to make sure this is ok?

frontend/client/src/screens/Settings.js
ibeckermayer: https://github.com/covidwatchorg/portal/issues/449

Issue #423, opened 2020-07-29
Cache control settings have not been set allowing the browser and proxies to cache content


whaber:
This potentially allows users of the same computer and/or proxies that may be between the browser and the portal server to access confidential information.

Ref on how to solve: https://cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html#web-content-caching

> Even after the session has been closed, it might be possible to access the private or sensitive data exchanged within the session through the web browser cache. Therefore, web applications must use restrictive cache directives for all the web traffic exchanged through HTTP and HTTPS, such as the `Cache-Control` and `Pragma HTTP headers`, and/or equivalent META tags on all or (at least) sensitive web pages.
>
> Independently of the cache policy defined by the web application, if caching web application contents is allowed, the session IDs must never be cached, so it is highly recommended to use the `Cache-Control: no-cache="Set-Cookie, Set-Cookie2"` directive, to allow web clients to cache everything except the session ID...

ibeckermayer:
Is this meant to be set in all responses? Firebase Cloud Functions/Auth may have our hands tied.

FYI, we're using Firebase's [out-of-the-box session management api](https://firebase.google.com/docs/auth/web/auth-state-persistence)

whaber:
Let's start with making this change:

`firebase.auth.Auth.Persistence.NONE`

Indicates that the state will only be stored in memory and will be cleared when the window or activity is refreshed.

ibeckermayer:
(Leaving a note for myself)

We're also manually saving the mobx store in sessionStorage, which should be removed as well:

```javascript
// Based on https://egghead.io/lessons/react-store-store-in-local-storage
if (sessionStorage.getItem('store')) {
  initialStore = JSON.parse(sessionStorage.getItem('store'))
}

const rootStore = Store.create({
  ...initialStore,
})

// Based on https://egghead.io/lessons/react-store-store-in-local-storage
onSnapshot(rootStore, (snapshot) => {
  sessionStorage.setItem('store', JSON.stringify(snapshot))
})
```

PR #446, opened 2020-08-03
Test Routes, Footer and Nav Bar

colbymorrison:
Builds on #325 and adds unit tests for the Footer and Nav Bar (rows 5, 9, and 14 of [the manual QA test sheet](https://docs.google.com/spreadsheets/d/1Gp38kggyQKlHOQLZ610TpGOg0cpsi3tQ5kdDQyjHWiY/edit#gid=0)) and integration tests for routes (rows 7 and 10 of the sheet).

### Usage
- Uses [Jest](https://jestjs.io/) as a testing framework and [Enzyme](https://enzymejs.github.io/enzyme/) as a testing utility to make testing for React easier. 

- After checking out this branch run
`npm install` 
`npm run test`

### Notes
- tests should be placed in the `__tests__` directory
- `test-setup.js` sets up the enzyme environment and is automatically ran before each test file. This is setup in `jest-config.js`
- What caused me a lot of trouble is mocking the store. For the footer/nav bar tests (or tests of any component that isn't App), we just create a store and manually pass it in as a prop. For testing `App.js` we use the store as is but mock all Firebase functionality. This required changing `auth.setPersistence` (line 23 of `store/index.js) to NONE for these test cases, or else there is an error. However, #423 suggests changing this to NONE for all cases, so maybe we should do that instead.

- Only some of the AddMember Modal tests from the previous PR work, I didn't look into them much. 

- Removes Logging from `src/store` files

### Updates after First Review
- Updated names of tests to be more readable:
    - Put all tests in a `describe` block
    - Changed instances of `test` to `it`, these are the exact same function, but allow for more human readable naming. For example, we have a test titled "it renders correctly"
- Added snapshot tests for Footer and NavBar
- Added the global `waitForComponentToPaint` function in `test-setup.js`. This is used by a few tests to avoid the "component not wrapped" warning.


frontend/client/__tests__/navbar.test.js
```diff
@@ -0,0 +1,56 @@
...
+import NavBar from '../src/components/NavBar'
+import { auth } from '../src/store/firebase'
+import { createStore } from '../src/store'
+
+beforeAll(() => {
+  rootStore.user.__update({ isSignedIn: true, isFirstTimeUser: false, firstName: 'testUser' })
+})
+
+// Mock Redirect to avoid router error
+jest.mock('react-router-dom', () => {
```
ibeckermayer, line 14: What is this doing?

frontend/client/__tests__/routes.test.js
```diff
@@ -0,0 +1,72 @@
...
+import CodeValidations from '../src/screens/CodeValidations'
+import Login from '../src/screens/Login'
+import reactRouter from 'react-router-dom'
+import ChangePasswordModal from '../src/components/ChangePasswordModal'
+
+// Make sure we don't call any Firebase functions
+jest.mock('../src/store/firebase')
+
+beforeAll(() => {
+  // Mock BrowserRouter as it overrides MemoryRouter used in test
```
ibeckermayer, line 16: Is this really "mocking" the BrowserRouter? Or is it just setting it to something else? My understanding is that mocking something requires a call to `jest.mock()`.

frontend/client/__tests__/routes.test.js
```diff
@@ -0,0 +1,72 @@
...
+test('404 for invalid link', () => {
+  const wrapper = mount(
+    <MemoryRouter initialEntries={['/random']}>
+      <App />
+    </MemoryRouter>
+  )
+  expect(wrapper.find(NotFound)).toHaveLength(1)
+})
+
+test('/ leads to /code_validations', () => {
```
ibeckermayer, line 31: How is the user being registered as "logged in", such that this test actually works? In `'non-auth routes'`, you start by calling `rootStore.user.__update({ isSignedIn: true, isAdmin: false, isFirstTimeUser: false })` in order to mock a log in, but here you don't.

frontend/client/__tests__/__snapshots__/addmembermodal.react.test.js.snap
```diff
@@ -0,0 +1,120 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
ibeckermayer, line 1: Has this been checked for accuracy? I believe Serena created this, do we know that it works properly?

frontend/client/__tests__/addmembermodal.react.test.js
```diff
@@ -0,0 +1,153 @@
+import { JSDOM } from 'jsdom'
+
+const dom = new JSDOM('<!doctype html><html><body></body></html>')
+global.window = dom
+global.document = dom.document
```
ibeckermayer, line 5: I think we can delete everything above this line

frontend/client/__tests__/addmembermodal.react.test.js
```diff
@@ -0,0 +1,153 @@
...
+
+import React from 'react'
+import Adapter from 'enzyme-adapter-react-16'
+import { configure, shallow, mount } from 'enzyme'
+import toJson from 'enzyme-to-json'
+import { rootStore } from '../src/store/model'
+import AddMemberModal from '../src/components/AddMemberModal'
+
+configure({ adapter: new Adapter() })
```
ibeckermayer, line 14: Do we need this line? Shouldn't it have already been called in `test-setup.js`?

ibeckermayer:
Is there some way that we're keeping track of which tests we've automated in the [Covid Watch Portal Manual QA Checklist](https://docs.google.com/spreadsheets/d/1Gp38kggyQKlHOQLZ610TpGOg0cpsi3tQ5kdDQyjHWiY/edit?pli=1#gid=0)?

frontend/client/__tests__/__snapshots__/addmembermodal.react.test.js.snap
colbymorrison: I left this file alone in this PR and updated it the test Manage Members PR!

frontend/client/__tests__/routes.test.js
colbymorrison: We just want BrowserRouter to not do anything or else it interferes with MemoryRouter used in the tests. Jest.mock is only needed if we need to watch what happens to the component during the test. I was just using mock as a general term for replacing the functionality of something. This idea taken from  https://medium.com/@antonybudianto/react-router-testing-with-jest-and-enzyme-17294fefd303

frontend/client/__tests__/addmembermodal.react.test.js
colbymorrison: See Manage Members test PR

frontend/client/__tests__/navbar.test.js
colbymorrison: This is mocking the Redirect component used in NavBar. If we don’t do this, an error is thrown than you can’t use Redirect outside of a Router. So, we just replace the Redirect component with an empty div

frontend/client/__tests__/addmembermodal.react.test.js
colbymorrison: See test manage members PR

frontend/client/__tests__/routes.test.js
colbymorrison: Oh I think this is supposed to test not logged in, as it’s checking if the Login page is displayed. We should change the title of the test.

colbymorrison:
Jesse added an extra column in the spreadsheet to keep track of which ones we’ve automated

colbymorrison:
Re: snapshot tests. Yes I think that’s a good idea, we can add it for sure.

frontend/client/__tests__/addmembermodal.react.test.js
```diff
@@ -0,0 +1,153 @@
+import { JSDOM } from 'jsdom'
```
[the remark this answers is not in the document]
colbymorrison: Will update in test manage Members PR

Issue #450, opened 2020-08-04
Reset password workflow logs user in with magic link without resetting password


jcolla-holla:
[Associated Asana task
](https://app.asana.com/0/1187032633259187/1187526409619473)

[Video of current behavior here](https://drive.google.com/file/d/1UWtMs6VdPAIWfmqkRlYgli_AaBRy2YvD/view?usp=sharing).  When doing reset password workflow, after clicking magic link user should have to enter new password or exit reset password workflow to login with current password.

PR #451, opened 2020-08-04
Fix "reset password workflow logs user in with magic link without resetting password"

ibeckermayer:
Fixes a bug where a user could request a password reset, click the magic link, get the supposed-to-be-un-dismissible reset modal, but then refresh the page and be logged in without ever actually resetting the pwd (see the video link in https://github.com/covidwatchorg/portal/issues/450)

Also fixes the NavBar weirdness (shown at the end of the video linked in https://github.com/covidwatchorg/portal/issues/450), so that at the points during the pwd reset where the user is technically logged in but hasn't completed the process, the navbar appears to them the same as if they were logged out.

closes https://github.com/covidwatchorg/portal/issues/450

### DIFF
--- a/frontend/client/src/store/index.js
+++ b/frontend/client/src/store/index.js
@@ -5,6 +5,7 @@ import {
   auth,
   db,
   SESSION,
+  NONE,
   createUserCallable,
   initiatePasswordRecoveryCallable,
   getVerificationCodeCallable,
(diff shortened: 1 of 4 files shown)


frontend/client/src/store/index.js
```diff
@@ -44,6 +44,13 @@ const createStore = (WrappedComponent) => {
                   isSignedIn: true,
                   signedInWithEmailLink: this.__signedInWithEmailLink,
                 })
+                // If this is a login triggered by a reset password email that the user clicked, set auth persistence to NONE
+                // so that they cannot escape the non-dismissable dialog and log in by refreshing the page
+                if (updatedUserDocumentSnapshot.data().passwordResetRequested && this.__signedInWithEmailLink) {
```
ibeckermayer, line 49: This is the section that prevents the refresh from logging the user in

PR #453, opened 2020-08-05
Test code validations

colbymorrison:
Depends on #446 

The new file in this PR is `__tests__/code_validations.test.js` All others are from #446.

Tests
- Can't generate a code without selecting a diagnosis type
- Can't input dates more than 14 days in the past or in the future
- Clicking the  "Generate Code" button calls the correct function in the backend

### Updates
- Fixed "An update to wrappedComponent..." warning
- Renamed tests to be more consistent


colbymorrison:
To fix the “An update to wrappedComponent...” Warning, see Line 36 of reset-password.test.js from the reset password test PR. That resolved this warning in other tests, I forgot to add it here.  The other warnings I’ve seen as well and not sure how to resolve so we can look into it!

PR #454, opened 2020-08-05
non-bootstrap mobile refactor for Login / NavBar / Footer

ibeckermayer:
### include-media
A potential direction for the mobile refactor that makes use of css media queries, rather than retrofitting the codebase with Bootstrap. 

I found a nice css library called [include-media](https://eduardoboucas.github.io/include-media) that makes the breakpoints easier to read and keep consistent, i.e. we can set the following in `_include-media.scss`
```scss
// _include-media.scss

$breakpoints: (
  'phone': 420px,
  'tablet': 768px,
  'desktop': 1024px,
) !default;
```
so that if we write
```scss
@include media('<=phone') {
    margin: 0 5px;
}
```
it compiles down to
```css
@media (max-width: 420px) {
    margin: 0 5px;
}
```

### Login
The login screen on mobile should look something like the following
Screen Shot 2020-08-05 at 2 34 24 PM

Rather than deal with trying to fit the big welcome blurb and picture of students hanging out with each other into this tiny interface, I elected to just set `display: none;` and eliminate them.

### Footer
The footer scss code is a bit convoluted, but it appears to get the job done. The key bit of trickery is
```scss
@include media('>phone') {
    flex-direction: row;
    ...
}
  @include media('<=phone') {
    flex-direction: column;
}
```
which ensures the footer stack horizontally on desktop, and vertically on mobile.

### NavBar
For the NavBar I made the call that it would be simplest to just remove the avatar/name/role section of the display (when logged in). So for desktop you should have something like this:
Screen Shot 2020-08-05 at 2 41 16 PM
which should become this on mobile:
Screen Shot 2020-08-05 at 2 42 54 PM

### To Test
The dev tools are an obvious place to start. I also found it useful to expose the app to the internet using `ngrok`
```
ngrok http 3000 -host-header="localhost:3000"
```
and check it out on my phone, since I've found that sometimes (often) there are discrepancies between the display in dev tools and things look on an actual mobile device.


leon-i:
Looks great, I think this definitely may be a more adaptive/easier way to make the portal mobile friendly. The breakpoint system with `@include` seems very easy to use.  
  
Login screen looks good, just think we need to center the "Covid Watch Portal" text and fix up the spacing. Also, I think we should still have a shortened version of the blurb somewhere with a link to the main site. Gonna post a mockup below.  
  
image
  
Only technical problem I encountered was with the logged-in navbar, it seems to be having the width issue I was experiencing while trying to use Bootstrap. Not sure if it's only a local issue. Here's a pic for reference:  
  
image  
  
As for moving forward, I think the final touches on the login page would be fixing up the toast and the forgot password modal, as seen below:  
  
image
  
image  
  
A separate issue can be made for these - for the toast I think we just need to center the text, and for the modal we just need to reduce its width a bit.

frontend/client/styles/components/navbar.scss
```diff
@@ -1,23 +1,39 @@
...
+    }
+    @include media('<=phone') {
+      height: 30px;
+    }
   }
 
   .menu-btn {
-    margin-right: 88px;
+    @include media('>phone') {
+      margin-right: 88px;
```
leon-i, line 32: Do we need the navbar to be pushed over for any particular reason? It kinda looks wonky to me like that, but if it's a design choice then nevermind.

frontend/client/styles/components/footer.scss
```diff
@@ -2,34 +2,61 @@
...
 #footer {
   display: flex;
+  // `margin-top: auto;` pins the footer to the bottom of the app, see https://stackoverflow.com/a/33924704
+  margin-top: auto;
 }
 
 .footer-not-login {
-  padding-right: 101px;
+  @include media('>phone') {
+    padding-right: 101px;
```
leon-i, line 11: Same thing as the navbar being pushed over.

frontend/client/styles/components/footer.scss
```diff
@@ -2,34 +2,61 @@
...
   align-items: center;
-  margin-left: 98px;
-  margin-right: 98px;
-
-
+  justify-content: space-between;
+  width: 100%;
+  @include media('>phone') {
+    flex-direction: row;
+    justify-content: space-between !important;
```
leon-i, line 22: Is there any reason why this needs to be important tagged and included here? Testing in chrome console showed no difference when removed.

frontend/client/styles/components/footer.scss
```diff
@@ -2,34 +2,61 @@
...
+    }
   }
-  
+
+  p,
   a {
     text-align: center;
-    margin-left: 50px;
+    @include media('>phone') {
+      margin-left: 50px !important;
```
leon-i, line 46: Do we need `!important` here? Testing w/o it showed no difference.

ibeckermayer:
> Looks great, I think this definitely may be a more adaptive/easier way to make the portal mobile friendly. The breakpoint system with `@include` seems very easy to use.

Agreed, let's go with this system.

> Login screen looks good, just think we need to center the "Covid Watch Portal" text and fix up the spacing. Also, I think we should still have a shortened version of the blurb somewhere with a link to the main site. Gonna post a mockup below.

I'm going to leave out the blurb for now, since that requires some copy work from the branding team. I'll include the "Learn more" link to the website.

> Only technical problem I encountered was with the logged-in navbar, it seems to be having the width issue I was experiencing while trying to use Bootstrap. Not sure if it's only a local issue. Here's a pic for reference:
>
> image

What's going on here is that the text/input boxes in CodeValidations.js are wider than the ordinary width of the screen, and so the screen is being shown zoomed out in order to accommodate them (the same was happening on the login page before I fixed it). So I'm going to ignore this for now and we can check that it gets fixed in my next PR where I optimize CodeValidations for mobile.

> As for moving forward, I think the final touches on the login page would be fixing up the toast and the forgot password modal, as seen below:
> 
> image
> 
> image
> 
> A separate issue can be made for these - for the toast I think we just need to center the text, and for the modal we just need to reduce its width a bit.

- https://github.com/covidwatchorg/portal/issues/460
- https://github.com/covidwatchorg/portal/issues/461

frontend/client/styles/components/navbar.scss
ibeckermayer: IIRC, we added that padding to try and set a maximum width for the app. According to git, you were the one who added that line ;-)

What aspect of it looks "wonky"? Can you post a screen shot? It could have something to do with different screen widths.

frontend/client/styles/components/footer.scss
ibeckermayer: See https://github.com/covidwatchorg/portal/pull/454#discussion_r467124279

frontend/client/styles/components/navbar.scss
ibeckermayer: This is a valid concern but not relevant to these changes, so let's push it off to another issue/PR if there's a better way to handle this

leon-i:
Merged #454 into `dev`.

PR #455, opened 2020-08-06
Test manage members

colbymorrison:
Depends on #446 . New files in this PR is `__tests__/manage_members.test.js`

### Tests
- Clicking the `Reset Password` link calls the reset password cloud function
- AddMemberModal closes correctly (Serena wrote this one)
- With valid input, AddMemberModal calls the create user Firebase function on submit (Serena wrote this one too!)
- With invalid input new user cannot be created. Invalid input tested is 1) Not all fields filled out and 2) All fields filled out but invalid email

### Notes
- I made one change to `ManageTeams.js`. The testing framework has a tough time managing `ref`s, so all the Toasts were breaking in the tests, even though the tests don't use or reference the Toast. Specifically, `confirmationToast.current` was always null. So I changed a line from `confirmationToast.current.show()` to ` confirmationToast.current && confirmationToast.current.show()`. This allows the tests to pass and doesn't change the functionality. Hope that is ok. If not, we can go another route.

### Updates
- Refactored Add Member Modal test to be more concise and renamed its extension from `.react.test.js` to `.test.js`
- Use global `waitForComponentToPaint` function instead of defining it locally


colbymorrison:
The add member modal tests are in addmembermodal.react.test.js, the file Serena submitted. I updated it in this PR (though I think I removed the snapshot test since it was failing, so we may want to re-add it)

PR #457, opened 2020-08-06
Test ResetPasswordModal and ChangePasswordModal

colbymorrison:
Depends on  #446. New file in this PR is `__tests__/reset_password.test.js`

### Tests
- With valid input, the `updatePassword` auth function is called
- Can't change password when the "New Password" and "Confirm Password" fields don't match
- New password can't be the same as old password (ResetPasswordModal only, see #416)

### Updates
- Use global `waitForComponentToPaint`
- More consistent test naming
- Fix new password == current password test


frontend/client/__tests__/reset-password.test.js
```diff
@@ -0,0 +1,143 @@
...
+describe('reset password', () => {
+  const waitForComponentToPaint = async (wrapper) => {
+    await act(async () => {
+      await new Promise((resolve) => setTimeout(resolve, 0))
+      wrapper.update()
+    })
+  }
+
+  test.each([
+    ['Reset Password Modal success', ResetPasswordModal, 'password1', 'password1'],
```
ibeckermayer, line 44: When I change `'password 1'` to `'password 2'` and vice versa, the tests still pass. Based on my experience with using `jest` for the firebase tests, this typically means that there's some asynchronous call that occurs, and `jest` just "jumps over it", reaches the end of the function, and says the test passed

frontend/client/__tests__/reset-password.test.js
```diff
@@ -0,0 +1,143 @@
...
+    const ModalWrapped = createStore(ResetPasswordModal)
+    const onSuccess = jest.fn()
+    const onFailure = jest.fn()
+    const wrapper = mount(<ModalWrapped onSuccess={onSuccess} onFailure={onFailure} />)
+
+    // Mock current password == new password error from firebase
+    auth.currentUser.reauthenticateWithCredential.mockImplementation(() => Promise.reject('expected error'))
+
+    const currentPassword = 'password'
+    const newPassword = 'password1'
```
ibeckermayer, line 107: How is this checking that "changing to current password is invalid", if the `newPassword !== currentPassword`?

frontend/client/__tests__/reset-password.test.js
colbymorrison: I think the way this is setup using test.each, either combination of password 1 / password 1 or password 1 / password 2 will work. But I believe the functionality is being tested.

frontend/client/__tests__/reset-password.test.js
colbymorrison: Good point, I realized this actually tests the current password is invalid case, since I mock reauthenticateWithCredential to throw an error. Whoops!

PR #459, opened 2020-08-06
Change auth state persistence to NONE

ibeckermayer:
closes https://github.com/covidwatchorg/portal/issues/423

Sets Firebase auth state-persistence to `NONE` and removes `sessionStorage` caching of the mobx `store` for non-development environments. We're keeping these for dev, since otherwise every time the webpack dev server hot-reloads you get logged out, which makes development extremely tedious.

To test:
- `npm run start:test # start the dev server with NODE_ENV=test`
- Login is admin@soylentgreen.com
- Refresh the page and see that you're logged out
- Ctrl-C to stop the dev server
- `npm run start:dev # start the dev server with NODE_ENV=development`
- Login is admin@soylentgreen.com
- Refresh the page and see that you're still logged in

### DIFF
--- a/frontend/client/src/store/index.js
+++ b/frontend/client/src/store/index.js
@@ -20,6 +20,8 @@ const createStore = (WrappedComponent) => {
   return class extends React.Component {
     constructor(props) {
       super(props)
+      // Set auth persistence to SESSION for development so that every change registered by the hot-reload dev server doesn't log you out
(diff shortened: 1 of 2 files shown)


jcolla-holla:
Followed the test flow you described (which makes sense to me based on the changes to the files) and works as expected.  Approved.

PR #462, opened 2020-08-07
Mobile refactor for code validations page

ibeckermayer:
Depends on https://github.com/covidwatchorg/portal/pull/454


jcolla-holla:
The width of the date input field seems to extend further than the width of the NavBar (and whole app on mobile - at least on an iPhone 5S screen).  This leads me to believe that the width of the input field should be made slightly less -- I've encountered this in other projects where the little date picker icon actually extends the width further than the CSS coded width.  I think that needs to be reduced.

Screen Shot 2020-08-07 at 2 42 58 PM

ibeckermayer:
Good catch, I also fixed the code generation on the dev infrastructure (and created this issue, since dev keeps on breaking on us), and fixed the "give out this code ASAP" text as well.

Issue #461, opened 2020-08-07
Fix forgot password modal on mobile


ibeckermayer:
Currently looks insane:
![image](https://user-images.githubusercontent.com/56734437/89568248-050c3e00-d7f1-11ea-880c-56befdfd2c42.png)

PR #469, opened 2020-08-10
Fix forgot pwd modal on mobile

ibeckermayer:
closes https://github.com/covidwatchorg/portal/issues/461


frontend/.prettierrc
```diff
@@ -2,5 +2,6 @@
   "trailingComma": "es5",
   "semi": false,
   "singleQuote": true,
-  "printWidth": 120
+  "printWidth": 120,
+  "tabWidth": 2
```
leon-i, line 6: What does adding `tabWith: 2` do?

ibeckermayer: It enforces that tabs are 2 spaces, rather than the default 4

ibeckermayer:
> Looks good! Not sure if it's just me - wondering if stylistically you think we should increase the padding a bit on the modals for the mobile site:
> 
> image
> 
> As seen above, they just seem a bit cramped. This also causes a few problems with the password successfully reset modal, as the clumped elements displace each other a bit and seem a bit off.
> 
> image

I agree with you that it still looks a little funky, though I would accept this as "good enough" for now, given the short notice and impending deadline. One thing we still need to sort out is whether we're going to get a professionally designed mobile interface from Vic for v5, or whether we're just going to keep winging it. In the meantime I think our best move is to just get to an acceptable mobile interface and then figure out what improvements we want to make (i.e. your suggestions).

All that said, if you're motivated to tinker around with these, you're free to do so and throw up a PR

Issue #464, opened 2020-08-08
Make text clickable for radio buttons in diagnosis code page


jcolla-holla:
To make it easier for users on their phone (and desktop actually) to click the radio buttons

PR #470, opened 2020-08-10
Make text clickable for radio buttons in diagnosis code page

ibeckermayer:
Gives each radio button an id and makes use of the `<label>` tag's `for` attribute to cause clicking on the label to select the radio button (https://stackoverflow.com/a/1146936)

closes https://github.com/covidwatchorg/portal/issues/464


frontend/client/styles/screens/code_validations.scss
```diff
@@ -91,6 +94,10 @@
 
         .col-2-sub-header {
           color: $darkMediumGray;
+          font-family: Montserrat;
```
colbymorrison, line 97: Can remove this line. Montserrat is set by default in `application.scss`

frontend/client/styles/screens/code_validations.scss
```diff
@@ -91,6 +94,10 @@
...
+          font-size: 14px;
+          font-weight: normal;
```
colbymorrison, line 99: Remove this line and line below (99-100), these are the defaults

ibeckermayer:
> See comments. Also, it is not due to this PR but the white circle in the radio button looks off-centered in a way it didn't before. Did this get changed in another PR?
> 
> radio

Browser/OS?

ibeckermayer:
> > See comments. Also, it is not due to this PR but the white circle in the radio button looks off-centered in a way it didn't before. Did this get changed in another PR?
> > radio
> 
> Browser/OS?

https://github.com/covidwatchorg/portal/issues/477

PR #471, opened 2020-08-10
My settings page mobile refactor

leon-i:
Refactored the my settings page to be mobile friendly using the include media library. Desktop site looks the same, mobile view shown below:  
  
image  
  
To-do:  
* Profile picture mobile modal fix  
* Change password mobile modal fix


leon-i:
Stacked the inputs in one row instead of 2:  

mobile2  
  
Also changed breakpoints for the grid to line up with the includemedia ones. For a larger overhaul, might need to do a separate issue/PR and convert stuff over. Do you want to move away from the material-ui grid system?

frontend/client/src/screens/Settings.js
```diff
@@ -15,6 +16,17 @@ import ResetPasswordModal from '../components/ResetPasswordModal'
 import CircularProgress from '@material-ui/core/CircularProgress'
 
 const MAXFILESIZE = 10 * 1024 * 1024
+const theme = createMuiTheme({
+  breakpoints: {
+    values: {
```
ibeckermayer, line 21: Good work, let's just add a comment explaining that these are set to correspond to the _include-media.scss breakpoints.

ibeckermayer:
Also please add issues for:

- Profile picture mobile modal fix
- Change password mobile modal fix

if they still need fixing (Change password looks ok to me)

frontend/client/src/screens/Settings.js
leon-i: Sounds good, I'll add it in

leon-i:
image
image
  
They both look kinda ok to me locally - do we want to just fix up the X button overlap for the pfp upload one?

PR #476, opened 2020-08-11
email input selected on page load

jcolla-holla:
Now the email field on the login screen is auto-selected when the user loads Portal.  It's a minor UX improvement so they don't have click on it to fill it out.  

Preview:

ezgif com-video-to-gif


.gitignore
```diff
@@ -1,3 +1,3 @@
 .DS_Store
 **node_modules/**
-.vscode
\ No newline at end of file
+.vscode
```
jcolla-holla, line 3: note: this is just tracking an addition and subtraction of a single line `package-lock.json`.  You can ignore this file as no change is being made here, effectively.

frontend/client/src/screens/Login.js
```diff
@@ -30,6 +30,10 @@ const SignInFormBase = observer(
       this.trySignInWithEmailLink()
     }
 
+    componentDidMount() {
```
jcolla-holla, line 33: The frontend tests are failing and I believe it because of the interaction between:

1) how the enzyme wrapper is being constructed
2) using `componentDidMount()`

Not totally sure why but I don't believe the frontend test errors are anything to be concerned with?  @colbymorrison thoughts?

leon-i: Have you tried using `React.createRef()`? Works very similar to what you're trying to do, but I think it plays nicer.  
  
Docs: https://reactjs.org/docs/refs-and-the-dom.html

colbymorrison: I encountered this error before. The issue is actually using `document.getElementById.` The tests use jsdom to imitate the DOM in the browser. However, the imitation is not perfect and `document.getElementById` always returns null in the test environment. See `frontend/client/__tests__/code_validations.test.js` Line 74-80 for how I got around this, but I think Leon is correct, the better approach is to use refs, I think they're the more React-y way to do this.

jcolla-holla: Updated using `createRef` instead.  Thx for the feedback!

leon-i:
Merged #476 into `dev`.

PR #487, opened 2020-08-13
Run workflow job based on files committed

### LINKED ISSUE #480: Make frontend and firebase tests auto-pass in CI if no changes have been made
Not super high priority but a helpful project that will require a somewhat creative solution:

Currently we require 5 actions to run before merging into dev (see `.github/workflows/build-and-test.yml`), which make up a firebase build/deploy/test and a frontend build/test. However for many PR's, one of those two sequences is unnecessary, per no changes having been made. I.e. often we're just working on the frontend, and so there's no reason to build/deploy/test the firebase cloud functions of that PR.

To solve this, we can design some system that checks we're running on a CI server (probably some environment variable), and then if so checks whether the current commit that's being tested has any diff from the current dev branch's `cloud-functions/` or `client` directories, and then based on that either triggers the full build/[deploy]/test or not.

One tricky aspect is that we still want to make each of these tests "required" tests in Github, and it doesn't give us programatic flexibility there (i.e. we can't say a test is required only if some condition is met). So the solution will need to account for that, probably by having a way for the actions that we decide don't need to run on some particular commit just return "success" immediately.

colbymorrison:
Closes #480 

GitHub Actions actually makes this not too bad to do using [outputs](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#outputs). 

We add another job in both workflows called `check-diff` that runs before all other jobs. `check-diff` determines if `frontend/` or `cloud-functions/` was altered by the previous commit. If one of the directories was altered, an output with its name is set  (see [here](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjobs_idoutputs) for a general example).  Later jobs check the outputs and only execute if the correct output was set. The fronted jobs will only run if `frontend/` or `config/` changed in the previous commit and the firebase jobs only run if `cloud-functions/` or `config/` changed in the previous commit. 

For example, in [this run](https://github.com/covidwatchorg/portal/actions/runs/206613171), I only changed the workflow file, so none of the frontend or cloud functions jobs ran. 

### Updates
`check-diff` is now its own custom action, that is used in both workflows.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -129,6 +129,7 @@ jobs:
   frontend-tests:
     name: Run Frontend Tests
     runs-on: ubuntu-latest
+    if: needs.check-diff.outputs.frontend == 'true'
     steps:
       - uses: actions/checkout@v2
       - name: Use Node.js 12.x
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -7,9 +7,33 @@ on:
     branches: [master, dev]
 
 jobs:
+  check-diff:
+    runs-on: ubuntu-latest
+    # Map a step output to a job output
+    outputs:
+      frontend: ${{ steps.check-diff.outputs.frontend }}
+      cloud-functions: ${{ steps.check-diff.outputs.cloud-functions }}
+
+    steps:
+      - name: Checkout master
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+
+      - id: check-diff
+        run : |
+           diff=$(git diff --name-only HEAD HEAD~1)
+           for dir in frontend cloud-functions; do
+              if echo $diff | grep -q "$dir\|config"; then
+                echo "::set-output name=$dir::true"
+              fi 
+           done
+
   build-frontend:
     name: Build Frontend App
     runs-on: ubuntu-latest
+    needs: check-diff
+    if: needs.check-diff.outputs.frontend == 'true'
     steps:
       - uses: actions/checkout@v2
       - name: Use Node.js 12.x
@@ -28,6 +52,8 @@ jobs:
   build-firebase:
     name: Build Firebase
     runs-on: ubuntu-latest
+    needs: check-diff
+    if: needs.check-diff.outputs.cloud-functions == 'true'
     steps:
       - uses: actions/checkout@v2
       - name: Install
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -9,25 +9,17 @@ on:
 jobs:
   check-diff:
     runs-on: ubuntu-latest
-    # Map a step output to a job output
+    # Map outputs from check-diff step to job outputs
     outputs:
       frontend: ${{ steps.check-diff.outputs.frontend }}
       cloud-functions: ${{ steps.check-diff.outputs.cloud-functions }}
-
     steps:
-      - name: Checkout master
+      - name: Checkout branch
         uses: actions/checkout@v2
         with:
           fetch-depth: 2
-
       - id: check-diff
-        run : |
-           diff=$(git diff --name-only HEAD HEAD~1)
-           for dir in frontend cloud-functions; do
-              if echo $diff | grep -q "$dir\|config"; then
-                echo "::set-output name=$dir::true"
-              fi 
-           done
+        uses: ./.github/actions/check-diff
 
   build-frontend:
     name: Build Frontend App
@@ -129,7 +121,7 @@ jobs:
   frontend-tests:
     name: Run Frontend Tests
     runs-on: ubuntu-latest
-    if: needs.check-diff.outputs.frontend == 'true'
+    needs: build-frontend
     steps:
       - uses: actions/checkout@v2
       - name: Use Node.js 12.x
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -8,9 +8,33 @@ on:
     branches: [dev, master]
 
 jobs:
+  check-diff:
+    runs-on: ubuntu-latest
+    # Map a step output to a job output
+    outputs:
+      frontend: ${{ steps.check-diff.outputs.frontend }}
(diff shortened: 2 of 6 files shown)


ibeckermayer:
> Closes #480
> 
> GitHub Actions actually makes this not too bad to do using [outputs](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#outputs).
> 
> We add another job in both workflows called `check-diff` that runs before all other jobs. `check-diff` determines if `frontend/` or `cloud-functions/` was altered by the previous commit. If one of the directories was altered, an output with its name is set (see [here](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjobs_idoutputs) for a general example). Later jobs check the outputs and only execute if the correct output was set. The fronted jobs will only run if `frontend/` or `config/` changed in the previous commit and the firebase jobs only run if `cloud-functions/` or `config/` changed in the previous commit.
> 
> For example, in [this run](https://github.com/covidwatchorg/portal/actions/runs/206613171), I only changed the workflow file, so none of the frontend or cloud functions jobs ran.

One issue this approach might cause is if those jobs don't run, but are considered "required" before merge by our branch protection rule, it might not allow us to merge (without an admin override).

One way we might solve that is to check within the tests "if CI, if <cloud-functions/frontend> not in check-diff output, pass-all-tests (or don't run tests, or some sort of jest hack)"

On the other hand, maybe branch protection rules are smart enough to account for this. May as well try this simpler approach first, and then if it doesn't work with the protection rules try this other one I just laid out.

colbymorrison:
That makes sense. Once this is approved we can see if it will let us merge. If not, we can rework.

.github/workflows/build-and-test.yml
```diff
@@ -7,9 +7,33 @@ on:
     branches: [master, dev]
 
 jobs:
+  check-diff:
+    runs-on: ubuntu-latest
+    # Map a step output to a job output
+    outputs:
+      frontend: ${{ steps.check-diff.outputs.frontend }}
+      cloud-functions: ${{ steps.check-diff.outputs.cloud-functions }}
+
```
ibeckermayer, line 16: Let's keep stylistically consistent with the other jobs by removing these empty lines

.github/workflows/build-and-test.yml
```diff
@@ -7,9 +7,33 @@ on:
...
+    steps:
+      - name: Checkout master
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+
+      - id: check-diff
+        run : |
+           diff=$(git diff --name-only HEAD HEAD~1)
```
ibeckermayer, line 25: Please add a comment explaining what this checks for

.github/workflows/build-and-test.yml
```diff
@@ -103,6 +129,7 @@ jobs:
   frontend-tests:
     name: Run Frontend Tests
     runs-on: ubuntu-latest
+    if: needs.check-diff.outputs.frontend == 'true'
```
ibeckermayer, line 132: I think this block needs a `needs: check-diff`? 

Alternately, we could make this block have a `needs: build-frontend`, since there's not much point in running the frontend tests if the app doesn't even build. And then in that case, the `needs: check-diff` in `build-frontend` will halt this one's execution as well.

.github/workflows/deploy.yml
```diff
@@ -8,9 +8,33 @@ on:
     branches: [dev, master]
 
 jobs:
+  check-diff:
```
ibeckermayer, line 11: Is there any way to share `check-diff` between both files?

.github/workflows/deploy.yml
```diff
@@ -8,9 +8,33 @@ on:
...
+    outputs:
+      frontend: ${{ steps.check-diff.outputs.frontend }}
+      cloud-functions: ${{ steps.check-diff.outputs.cloud-functions }}
+
+    steps:
+      - name: Checkout master
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+
```
ibeckermayer, line 23: Could remove the empty lines within jobs in this file too, for the sake of stylistic consistency between the two.

.github/workflows/build-and-test.yml
```diff
@@ -7,9 +7,33 @@ on:
...
+    # Map a step output to a job output
+    outputs:
+      frontend: ${{ steps.check-diff.outputs.frontend }}
+      cloud-functions: ${{ steps.check-diff.outputs.cloud-functions }}
+
+    steps:
+      - name: Checkout master
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
```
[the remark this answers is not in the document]
colbymorrison: This loads the previous 2 commits into the GH workflow environment. Only one commit is loaded by default. This is needed or else the `git diff` line below won't work. See https://github.com/actions/checkout for more info.

.github/workflows/build-and-test.yml
```diff
@@ -7,9 +7,33 @@ on:
...
 jobs:
+  check-diff:
+    runs-on: ubuntu-latest
+    # Map a step output to a job output
+    outputs:
+      frontend: ${{ steps.check-diff.outputs.frontend }}
+      cloud-functions: ${{ steps.check-diff.outputs.cloud-functions }}
+
+    steps:
+      - name: Checkout master
```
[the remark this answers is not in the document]
colbymorrison: Yeah, it's whatever branch was just pushed. Will update the title.

.github/workflows/build-and-test.yml
colbymorrison: I like that idea! Will update with `needs: build-frontend`

.github/workflows/deploy.yml
colbymorrison: We could try to make it is own action that is called by both files. I can give that a shot.

colbymorrison:
Yep, something is still off for sure. I’ll take a look when I get internet back at my house (commenting from mobile).

ibeckermayer:
Based on DM's w/ @colbymorrison: It seems as though all of our tinkering threw GitHub actions into a weird state. Closing this, replaced by https://github.com/covidwatchorg/portal/pull/516/.

PR #488, opened 2020-08-13
Temporary Manage Members Mobile Screen for Friday

leon-i:
Members grid hidden on mobile until further refactoring can be made. Still able to add members, displays message to use desktop if you want to see members grid. Pic below:  
  
image  
  
Also added link to manage members screen back to mobile nav menu.

### DIFF
--- a/frontend/client/styles/screens/manage_teams.scss
+++ b/frontend/client/styles/screens/manage_teams.scss
@@ -18,11 +18,10 @@
 }
 
 .mobile-msg {
-  display: none;
+  align-self: center;
 
-  @include media('<=phone') {
-    display: block;
-    align-self: center;
+  @include media('>phone') {
+    display: none;
   }
 }
 
--- a/frontend/client/styles/screens/manage_teams.scss
+++ b/frontend/client/styles/screens/manage_teams.scss
@@ -1,8 +1,13 @@
 @import '../color.scss';
+@import '../include-media';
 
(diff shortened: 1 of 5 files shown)


frontend/client/styles/screens/manage_teams.scss
```diff
@@ -12,11 +17,25 @@
   cursor: pointer;
 }
 
+.mobile-msg {
+  display: none;
```
ibeckermayer, line 21: Nitpick here:

Let's make this
```scss
@include media('>phone') {
    display: none;
}
```

As a general rule I think its much easier to read this conditional scss code when it's explicit about what's for mobile and what's for desktop. I've found when I don't write the code this way, I commonly start mixing the paradigm up (i.e. some classes the default is mobile, while others the default is desktop), and code integrity degrades over time.

leon-i: Fixed, no prob - makes sense

leon-i:
Fixed role select overflow, adjusted the height of the modal on mobile to better fit because it was overflowing a bit on smaller mobile devices (iPhone 5/5s). Pics below:  
  
Iphone X:  
image  
  
Iphone 5/5s:  
image

leon-i:
Fixed it so the Role select dropdown arrow now appears on smaller mobile devices

leon-i:
Merged #488 into `dev`.

Issue #479, opened 2020-08-12
Portal: Unable to click on the manage member button after clicking on the "Soylent Green" button.


ibeckermayer:
Asana task [here](https://app.asana.com/0/1188301658055194/1188621109875702).

Pre-condition:
User must be logged in with valid credentials

Step to reproduce:
1. Select "Manage Members" button from hamburger menu.
2. Click on the "Soylent Green" button to navigate to home page.
3. Again click on the hamburger menu and click on the "Manage Members" button.

Actual result:
Unable to open the Manage Member option.
Expected result:
Manage Member should open.

Platform: Web.
Browser: Chrome, Firefox, Edge

Please go through the video link for reference:
https://asrivastava-lwolf.tinytake.com/tt/NDUzMjE2OV8xNDM2MzYzNQ

PR #492, opened 2020-08-13
Fix navigation menu

ibeckermayer:
closes https://github.com/covidwatchorg/portal/issues/479

See the bug in [this video](https://asrivastava-lwolf.tinytake.com/tt/NDUzMjE2OV8xNDM2MzYzNQ) and see that this updated design fixes it. FTR, a similar bug could be triggered by clicking the image avatar to navigate to the settings page.

### DIFF
--- a/frontend/client/src/components/NavBar.js
+++ b/frontend/client/src/components/NavBar.js
@@ -1,5 +1,6 @@
 import React, { useState } from 'react'
-import { Redirect, Link } from 'react-router-dom'
+import { Link, useHistory } from 'react-router-dom'
+import * as ROUTES from '../constants/routes'
 import IconButton from '@material-ui/core/IconButton'
 import MenuItem from '@material-ui/core/MenuItem'
 import Menu from '@material-ui/core/Menu'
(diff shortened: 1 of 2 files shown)


frontend/client/src/components/NavBar.js
```diff
@@ -25,10 +26,18 @@ const NavBarBase = observer((props) => {
 
   const onClickMenuItem = (num) => {
     setAnchorEl(null)
-    if (num === 4) {
+    if (num === 0) {
```
leon-i, line 29: Have you considered using a switch statement here instead? Seems like the place for it.

jcolla-holla: I just experimented w switch statements and it would look like the below.  Not sure there is much to be gained by shifting the syntax to switch statements.

```javascript
const onClickMenuItem = (num) => {
    setAnchorEl(null)
    switch(num) {
      case 0:
        history.push(ROUTES.CODE_VALIDATIONS)
        break
      case 1:
        history.push(ROUTES.MANAGE_MEMBERS)
        break
      case 2: 
        history.push(ROUTES.BRANDING)
        break
      case 3:
        history.push(ROUTES.SETTINGS)
        break
      default:
        props.store.signOut()
        break
    }
}
```

ibeckermayer: It's just another way to express the same thing

ibeckermayer:
> Seems to fix the bug, but I'm wondering if it would be better to just use `Link` redux components instead for the menu items (or wrap the menu items in `Link` components)? Might be better to just convert it over now to save future band-aiding.

In principle makes sense to just make these "Link" components, though we're not using redux so I don't think those specifically would work.

PR #493, opened 2020-08-13
prevent name change save of one char

### LINKED ISSUE #490: User can delete and continue without setting any First and Last name from settings.
Asana link [here](https://app.asana.com/0/1188301658055194/1188824083782177/f)

jcolla-holla:
closes #490 

Note that the user is able to delete all characters from the `firstName` and `lastName` fields but the store is only altered if the field has one or more characters.  Meaning that a user can remove all characters, see an error message, but the name that is saved is a single character if they navigate away from the page.

I played with a few options like forcing a single character to remain in the input fields but this seemed like a bad UX if the user for some reason wants to delete all characters and start over, it prevents them from doing so and is a little jerky.


leon-i:
Merged #493 into `dev`.

Issue #435, opened 2020-07-30
It sends empty symptomDate even if it was selected during the code generation


jcolla-holla:
Reported by Pavlo...

It sends empty `symptomDate` even if it was selected during the code generation:

To reproduce
Generate a code with the date
Press Reset Code and Form
Try to select the current date
Generate the code
If you try to verify the code, the server information won't the the selected date during the step 3.
It is also visible on the UI, when the date is the selected "but not really" -- the day number is gray-ish

Issue #410, opened 2020-07-27
Calendar input api not working on Safari


ibeckermayer:
On Brave, the "Symptom Onset Date" is the inbuilt calendar input element:
<img width="541" alt="Screen Shot 2020-07-27 at 9 18 02 AM" src="https://user-images.githubusercontent.com/13578537/88546259-1c297f80-cfea-11ea-9990-1925e76f01d7.png">

However on Safari, users can input whatever they want into that box
<img width="551" alt="Screen Shot 2020-07-27 at 9 14 09 AM" src="https://user-images.githubusercontent.com/13578537/88546124-ef756800-cfe9-11ea-95a9-24306bc0e6dc.png">

<img width="559" alt="Screen Shot 2020-07-27 at 9 17 27 AM" src="https://user-images.githubusercontent.com/13578537/88546196-074cec00-cfea-11ea-95f0-8f7f96e3e36a.png">

colbymorrison:
Safari doesn't support a date picker :( 

https://stackoverflow.com/questions/35682138/html5-date-picker-doesnt-show-on-safari

Issue #489, opened 2020-08-13
No validation message is displayed and code is generated when user is deleting month or date section


jcolla-holla:
No validation message is displayed and code is generated when user is deleting month or date section.

Asana link [here](https://app.asana.com/0/1188301658055194/1188793474409842/f) with more details on how to reproduce.  If you need access to Asana let me know.

colbymorrison:
When the user deletes a month or date section, the date is interpreted as an empty string. We allow a code to be generated with just a diagnosis selection and no date, which I think is correct. So for the system, this is the same case (diagnosis selected and no date). 

So, if we should leave the date as optional what should the semantics be in this case?

PR #494, opened 2020-08-13
Custom Safari desktop date picker

### LINKED ISSUE #491: No prefilled placeholder text “dd-mm-2020” appears on date picker in Safari browser
No prefilled placeholder text “dd-mm-2020” appears on date picker field on “Diagnosis verification Codes” page in Safari browser

Asana [link here](https://app.asana.com/0/1188301658055194/1188835089164417/f).  If you need Asana access let me know.

jcolla-holla:
closes #491 #410 

Safari desktop [doesn't support a calendar picker](https://stackoverflow.com/questions/35682138/html5-date-picker-doesnt-show-on-safari) unfortunately (lame).

More robust solution is to populate a calendar picker no matter the browser, which is done here using [ReactDatePicker](https://reactdatepicker.com/).  I tested the whole workflow on Safari and all works as expected (same as Chrome).

Note that Safari mobile has a much better date picker UX.  I was concerned about this as it seems many users will be on mobile and Safari is the default browser for iPhones.  No need to adjust for this, thankfully (see screenshot below from my own iPhone on Safari):

IMG_A7E7458123B1-1

### DIFF
--- a/frontend/client/src/screens/CodeValidations.js
+++ b/frontend/client/src/screens/CodeValidations.js
@@ -199,6 +199,7 @@ const CodeValidationsBase = observer((props) => {
               min={getFourteenDaysAgoString()}
               max={getTodayString()}
               onChange={handleDate}
+              placeholder="mm-dd-yyyy"
             ></input>
           </form>
           <p className="date-desc">Time zone set to {getDefaultTimezoneString()}</p>
--- a/frontend/client/src/screens/CodeValidations.js
+++ b/frontend/client/src/screens/CodeValidations.js
@@ -11,8 +11,6 @@ import DatePicker from 'react-datepicker'
 
 import {
   getOneHourAheadDisplayString,
-  moreThanFourteenDaysAgo,
-  dateInFuture,
   getDefaultTimezoneString,
   localStringToZeroUTCOffsetString,
   toDashSeperatedYYYYMMDDString,
@@ -64,7 +62,8 @@ const CodeValidationsBase = observer((props) => {
     try {
       let code = await props.store.getVerificationCode({
         testType: testType,
-        symptomDate: symptomDateYYYYMMDD === '' ? symptomDateYYYYMMDD : localStringToZeroUTCOffsetString(symptomDateYYYYMMDD),
+        symptomDate:
+          symptomDateYYYYMMDD === '' ? symptomDateYYYYMMDD : localStringToZeroUTCOffsetString(symptomDateYYYYMMDD),
       })
       setCode(code.data.split('').join(''))
       codeTimeStamp()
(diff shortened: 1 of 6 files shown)


frontend/client/src/screens/CodeValidations.js
```diff
@@ -199,12 +213,19 @@ const CodeValidationsBase = observer((props) => {
...
               max={getTodayString()}
               onChange={handleDate}
+              placeholder="mm-dd-yyyy"
             ></input>
           </form>
           <p className="date-desc">Time zone set to {getDefaultTimezoneString()}</p>
         </div>
       </div>
 
+      {/* does this render a calendar picker upon clicking?  or a bunch of numbers vertically? */}
```
jcolla-holla, line 223: I want to see if this is occurring on other machines/browser versions.  Our React version 16.x doesn't seem to conflict w the [ReactDatePicker docs](https://reactdatepicker.com/).

But a strange behavior is occurring that I couldn't figure out exactly why.  See below.  Is this happening to you all also?  Note this is happening on all browsers I tried: Chrome, Safari, Brave.

Screen Shot 2020-08-13 at 4 32 20 PM

colbymorrison:
I think replacing the <input DatePicker with a package such as `react-datepicker` is the right way to go. For us, this resolves the Safari issue, and will likely resolve #435, and possibly #489.

frontend/client/src/screens/CodeValidations.js
```diff
@@ -124,6 +136,18 @@ const CodeValidationsBase = observer((props) => {
...
   }
 
+  let datePicker = (
+    <DatePicker
+      id="date-picker"
+      className={symptomDateYYYYMMDD ? 'with-value' : 'no-value'}
+      selected={symptomDateObject}
+      minDate={getFourteenDaysAgoDate()}
+      maxDate={new Date()}
+      onChange={(date) => handleDate(date)}
```
colbymorrison, line 146: This can just be onChange={handleDate}

frontend/client/src/screens/CodeValidations.js
```diff
@@ -79,36 +85,42 @@ const CodeValidationsBase = observer((props) => {
...
-    e.target.classList.remove('no-value')
+  const handleDate = (date) => {
+    setSymptomDateObject(date)
+    const selectedDate = toDashSeperatedYYYYMMDDString(date)
+    document.getElementById('date-picker').classList.add('with-value')
+    document.getElementById('date-picker').classList.remove('no-value')
 
     // does not allow symptomDate in state to be set if date selected is more than 14 days ago or in the future
-    if (moreThanFourteenDaysAgo(e.target.value)) {
+    if (moreThanFourteenDaysAgo(selectedDate)) {
```
colbymorrison, line 95: The `DatePicker` object does not allow a date outside its range to be selected. So no date passed to this function will be invalid. So we can delete these lines (95-109) and always run what is currently the else case. I think we can also delete the `dateInFuture` and `moreThanFourteenDaysAgo` functions from `time.js`, which is also good. 

In this recording: I put `console.log(date)` at Line 89. Note how for invalid dates, they are immediately reset to a valid date and not logged.  Also note the uncaught error when the date is cleared. 
vals1

jcolla-holla: Yup!  I was just about to commit this deletion myself!

colbymorrison:
I think this closes #494 as well. The picker forces a valid date to be set.

jcolla-holla:
> Requires some frontend tests to be updated before merging.
> 
> Additionally, we should make sure that this doesn't undo [this recently-merged fix](https://github.com/covidwatchorg/portal/pull/496).

@ibeckermayer frontend tests refactored and passing (Colby and I reviewed these together). Re: the other branch, I believe this should be ok to merge as it will just overwrite what's there -- no merge conflict issues are highlighted by Github.  

Once you approve feel free to merge and we can have the QA folks try to break it in staging.

ibeckermayer:
> @ibeckermayer frontend tests refactored and passing (Colby and I reviewed these together). Re: the other branch, I believe this should be ok to merge as it will just overwrite what's there -- no merge conflict issues are highlighted by Github.

I meant we should make sure it doesn't undo the functionality of the fix. I just checked and everything is running smoothly.

PR #498, opened 2020-08-14
Update VS urls

ibeckermayer:
Updating the verification server urls to distinguish between dev and prod. All infrastructures point to https://dev.adminapi.verification.covidwatch.org, except for prod which points to https://adminapi.verification.covidwatch.org.

Hardcodes the url strings rather than using environment variables, since they're not secret.

Also updates the test config to use the dev VS url and grab the API key from the env, in preparation for future integration test development.

NOTE-to-self: after this is merged would be a good time to go through our Github secrets and clean up some cruft.


ibeckermayer:
> Just to verify, once this is merged then any pings to our production server are the real, 'live' production server, yes? And we should be more cautious with testing those esp given the 128 API rate limit.

Yeah that's my understanding

Issue #295, opened 2020-06-23
PendingOperationButton bug


ibeckermayer:
If a user is logged in as a first time user such that the `ChangePasswordModal` automatically pops up, and they try to go to an alternative route (say `/code_validations`) causing them to automatically be redirected to back to login page with the `ChangePasswordModal` popped up, and then they change their password, we get the following warning in the console:

```
react-dom.development.js:88 Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
    in PendingOperationButton (created by ChangePasswordModalBase)
    in form (created by ChangePasswordModalBase)
```

fordacious:
Unmounted component means it has been deleted, or is not being rendered. Generally this is a data flow issue.

fordacious:
Based on the description I'm guessing some state change is going on in the old one referencing components that no longer exist.

colbymorrison:
We should try and resolve this.

PR #501, opened 2020-08-14
PendingOperationButton memory leak fix

ibeckermayer:
closes https://github.com/covidwatchorg/portal/issues/295

Inspired by https://dev.to/otamnitram/react-useeffect-cleanup-how-and-when-to-use-it-2hbm

To check:
Previously `reset-password.test.js` and `code_validations.test.js` were throwing
```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
```
but that should no longer be the case


frontend/client/src/components/PendingOperationButton.js
```diff
@@ -18,20 +18,39 @@ const PendingOperationButton = (props) => {
...
   const disabled = props.disabled || false
 
-  const startOperation = async () => {
-    setIsOperationPending(true)
+  // Flag to keep track of whether component is mounted. We check this flag in runOperation before setting the button's state,
+  // in order to prevent cases where we mistakenly try to update the state when the button is no longer mounted, which was
+  // happening relatively frequently in i.e. automatic-modal-closes and triggering the following warning:
+  // "Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application.
+  // To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function."
+  var componentIsMounted = false
```
colbymorrison, line 26: Can this be part of the state? I.e. change to `const [componentIsMounted, setComponentIsMounted] = useState(false)`

ibeckermayer:
> This breaks login for me, so I think there's still a bug somewhere
> Screen Capture_Navigator_20200814152848

Lol, yeah this isn't good... Broken for me too -- converting to draft and will dig in further.

frontend/client/src/components/PendingOperationButton.js
ibeckermayer: Why should it be?

PR #523, opened 2020-08-18
Add password strength meter

### LINKED ISSUE #247: Add password strength meter
When the user is setting their password, we should give them an estimate of the strength of the password, operationalized as the number of guesses needed to crack the password. We can use a standard password strength meter like [zxcvbn](https://github.com/dropbox/zxcvbn), which is included in [this React component](https://www.npmjs.com/package/react-password-strength).

Follow-up to #214.

For reference, I really like Slack's password strength meter:
Slack's password strength meter

veeara282:
Closes #247

Add a password strength meter to the "New password" fields in `ChangePasswordModal` and `ResetPasswordModal` using a `PasswordStrengthModalInput`, which composes `ModalInput`. This component will always display validation text, even when the password is valid.

Password strength indicator colors:
- Weak/invalid: `#f05452` (tangerine)
- Okay: `#43c4d9` (aqua)
- Good: `#388ec5`
- Great: `#2c58b1` (bluejay)

TODO:
- [x] Allow `{Reset,Change}PasswordModal` to read the validation result (true/false) of the `PasswordStrengthModalInput`
- [x] Color shouldn't be red when the password is valid

### DIFF
--- a/frontend/client/__tests__/reset-password.test.js
+++ b/frontend/client/__tests__/reset-password.test.js
@@ -68,7 +68,7 @@ describe('reset password', () => {
     }
     wrapper
       .find('#new-password')
-      .at(1)
+      .at(2)
       .simulate('change', {
         target: {
           name: 'new-password',
--- a/frontend/client/__tests__/reset-password.test.js
+++ b/frontend/client/__tests__/reset-password.test.js
@@ -33,14 +33,14 @@ jest.mock('../src/store/firebase', () => {
 
 describe('reset password', () => {
   it.each([
-    ['Changes password succesfully (ResetPasswordModal)', ResetPasswordModal, 'password1', 'password1'],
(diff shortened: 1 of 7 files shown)


aidan-fitz:
We should also consider adding some default and personalized strings to the `user_inputs` argument.

> The optional `user_inputs` argument is an array of strings that zxcvbn will treat as an extra dictionary. This can be whatever list of strings you like, but is meant for user inputs from other fields of the form, like name and email. That way a password that includes a user's personal information can be heavily penalized. This list is also good for site-specific vocabulary — Acme Brick Co. might want to include ['acme', 'brick', 'acmebrick', etc].

ibeckermayer (commented):
Ignore, I'm just clearing out old review requests in my GitHub account.

PR #528, opened 2020-08-19
org-specific-analytics

jcolla-holla:
No real code yet, just hashing out some ideas for feedback and to think through this a bit after reading through some documentation/watching some firebase analytics tutorials.  The goal we have is to easily view the number of verification codes generated by users with a specific `organizationID` in Portal.

Unfortunately, that doesn't appear possible out of the box the way that `getVerificationCode` "detailed logs" in Firebase currently (see image below).

Screen Shot 2020-08-19 at 3 51 09 PM

That being the case, there appear to be several options, the simplest of which seems to be just creating an additional [log event](https://firebase.google.com/docs/analytics/events) whenever `generateVerificationCode` is invoked and use the frontend state to pass in the desired data:

```javascript
analytics.logEvent('getVerificationCode_w_org_ID', { organizationID: props.store.data.user.organizationID, organizationName: props.store.data.organization.name} );
```

Then -- based on some small experiments I was doing to verify this -- I believe we can query for events `getVerificationCode_w_org_ID` and specify an `organizationID` and/or `organizationName` in the Firebase logs so that this solution can be used to query for ANY team beyond just UofA going forward.

Notes:
- [docs for query syntax in firebase logs](https://cloud.google.com/logging/docs/view/query-library#gae-filters)
- another potential route - [extending Google Analytics w Firebase functions](https://firebase.google.com/docs/analytics/extend-with-functions).  Ultimately Google Analytics and Firebase are two separate services, so the benefit of using log events is that these are logged in the Firebase logs (where the current cloud functions are all logged by default).  Best to keep everything related to cloud function invocations in one single place in my view.


ibeckermayer:
> Unfortunately, that doesn't appear possible out of the box the way that `getVerificationCode` "detailed logs" in Firebase currently (see image below).
> 
> Screen Shot 2020-08-19 at 3 51 09 PM

What's this screenshot from?

ibeckermayer:
> another potential route - [extending Google Analytics w Firebase functions](https://firebase.google.com/docs/analytics/extend-with-functions).  Ultimately Google Analytics and Firebase are two separate services, so the benefit of using log events is that these are logged in the Firebase logs (where the current cloud functions are all logged by default).  Best to keep everything related to cloud function invocations in one single place in my view.

I don't see any reason we need this, though I could be missing something...

frontend/client/src/screens/CodeValidations.js
```diff
@@ -70,6 +70,15 @@ const CodeValidationsBase = observer((props) => {
       setNeedsReset(true)
       updateButtonDisabled()
       countdown()
+
+      // WORK IN PROGRESS, notes:
+      // need to import analytics?  and make analytics testable in dev by removing conditional in firebase.js and resolve local error caused by removing it
```
ibeckermayer, line 75: > need to import analytics?

We can attach it to the store as a property called `analytics`, and then access it anywhere in the app via `props.store.analytics`.

> and make analytics testable in dev by removing conditional in firebase.js and resolve local error caused by removing it

Yeah, just remove the conditional and see what happens on infrastructure where analytics isn't enabled.

frontend/client/src/screens/CodeValidations.js
```diff
@@ -70,6 +70,15 @@ const CodeValidationsBase = observer((props) => {
...
+      // log a new Firebase event to count the number of codes generated by organization
+      
+      // analytics.logEvent('getVerificationCode_w_org_ID', {
```
ibeckermayer, line 78: Let's make this a bit simpler:

```javascript
analytics.logEvent('verificationCodeGenerated', {
    organizationID: props.store.data.user.organizationID,
})
```

jcolla-holla:
Pushed some code.  Addressing some of your points @ibeckermayer :

- the source of the screenshot of the production logs for `getVerificationCode` can be found [here](https://console.cloud.google.com/logs/viewer?project=permission-portal&authuser=0&resource=cloud_function%2Ffunction_name%2FgetVerificationCode%2Fregion%2Fus-central1&minLogLevel=0&expandAll=false&timestamp=2020-08-19T22:12:25.007000000Z&customFacets=&limitCustomFacetWidth=true&dateRangeStart=2020-08-19T21:12:25.258Z&dateRangeEnd=2020-08-19T22:12:25.258Z&interval=PT1H)
- I think including the name of the organization in the event is helpful so the user doesn't need to cross reference an ID with the name in Firebase in order to do some simple querying/data exploration

I think that these logged events will show up [here](https://console.firebase.google.com/u/0/project/permission-portal-dev/analytics/app/web:NWUzNjA0ZDAtMjViZC00MmI2LWFmYmItNTA2ZjczOGQ1N2Fj/events/~2Foverview%3Ft%3D1597990409543&fpn%3D885750041965&swu%3D1&sgu%3D1&sus%3Dupgraded&params%3D_r..layout.pageNumber%253D0%2526_u.dateOption%253DthisWeekSunToToday&cs%3Dapp.m.events.overview&g%3D1) (this is for Firebase project `permission-portal-dev`).  Right now for me the Firebase message in the UI says to wait up to 24 hours for the SDK to propagate. ?

Frontend tests - right now this PR is generating this error IDB requires a browser environment`, that I think is related to calling `const analytics = app.analytics()` in `firebase.js` without the conditional of only doing so in production.  My goal in removing the conditional is to test the event logging in Firebase dev/staging also before pushing to prod.  [This may offer solutions ](https://github.com/firebase/firebase-js-sdk/issues/2336#issuecomment-550090914)to address this error, requires some messing with Jest.  

Alternatively, can just adjust the conditional and exporting so `analytics` is defined when exported in test/dev env?

frontend/client/src/store/index.js
```diff
@@ -32,6 +33,8 @@ const createStore = (WrappedComponent) => {
       this.__signedInWithEmailLink = false // firebase doesn't tell us this so we need to track it ourself
       this.__authStateListener = auth.onAuthStateChanged(async (user) => {
         if (user) {
+          // make analytics accessible throughout app
+          this.__analytics = analytics
```
ibeckermayer, line 37: Let's move this up with the other class variables, since it doesn't need data from the user login.

Also we should name it just `this.analytics`. The `__` is a convention (from python) for defining "private" properties -- but since this property is meant to be used throughout the app (like `this.store`), it shouldn't be denoted as "private".

frontend/client/src/store/firebase.js
```diff
@@ -9,6 +9,7 @@ import * as firebaseConfigDev from '../config/firebase.config.dev'
 import * as firebaseConfigTest from '../config/firebase.config.test'
 import * as firebaseConfigProd from '../config/firebase.config.prod'
 import * as firebaseConfigStaging from '../config/firebase.config.staging'
+import { process } from 'mobx-state-tree'
```
ibeckermayer, line 12: Is this necessary?

frontend/client/src/screens/CodeValidations.js
```diff
@@ -73,6 +73,10 @@ const CodeValidationsBase = observer((props) => {
       setNeedsReset(true)
       updateButtonDisabled()
       countdown()
+      props.store.__analytics.logEvent('verificationCodeGenerated', {
```
ibeckermayer, line 76: Should become `props.store.analytics`

PR #532, opened 2020-08-29
Update Babel Config to Support Internet Explorer

colbymorrison:
Updates Babel and Webpack configs. They were a bit weird before so it was good we revisited them.

### Previously
- We configured babel in `webpack.common.js` for web  and `babel.config.js` for testing, this was awkard.
- For web we had babel target Node. I believe this meant that our js code was transpiled to run on Node, which is not what we want in any case, we want it to be transpiled for a browser! (Node is based on the Chrome runtime, so it worked just fine, but broke on IE as we saw)
- We used the `babel-polyfill` package to add our polyfills. This copied the entire `babel-polyfill` package to `dist/polyfill.js`.

### Now
- Only configure babel in `babel.config.js`. If `NODE_ENV == test`, target Node. Otherwise, target broswerslist's ["default" query](https://github.com/browserslist/browserslist#best-practices). Babel now reports it is targeting these browsers:
DeepinScreenshot_select-area_20200829124559

- Babel-polyfill is now [deprecated](https://babeljs.io/docs/en/7.9.0/babel-polyfill). So, I followed their instructions and installed [corejs@3](https://github.com/zloirock/core-js) explicitly. I added the `useBuiltIns` option in Babel ~~with the `usage` flag. This adds polyfills as imports to the transpiled js files~~ with the `entry` flag. This required placing some `imports` at the top of `client/index.js`. See the [docs](https://babeljs.io/docs/en/7.9.0/babel-preset-env#usebuiltins) for a better explanation of all this. ~~I couldn't get the `entry` flag to work, but it seems `usage` is more efficient anyway? (this confused me a bit)~~ Now, after `npm run build`, there is no `polyfill.js` file in `dist/`~~, so `dist/` is about a MB smaller since we only are using the polyfills we need, not all of them :)~~

-~~NOTE: IE11 still does not work because we are using Mbox 5+. See [here](https://mobx.js.org/README.html) and [here](https://babeljs.io/docs/en/learn#proxies) for details, this puts us at an impass for IE support.~~

### Internet Explorer Support

- Downgrades Mobx to version 4, version 5 is not comparable with IE ([docs](https://mobx.js.org/README.html)). The upcoming version 6 will be apparently, so we should keep an eye out for that. 
- In addition to the changes above, we ignore the `@fvilers/disable-react-devtools` package in the webpack config, so it is also transpiled by Babel. It causes an error in IE without being transpiled.
- We add the [cross-fetch](https://github.com/lquixada/cross-fetch) polyfills to `client/index.js`. This makes the cloud functions work in IE11. See [firebase docs](https://firebase.google.com/support/guides/environments_js-sdk#optional-polyfills).


ibeckermayer:
It looks like the frontend tests are failing

frontend/client/__tests__/__snapshots__/footer.test.js.snap
```diff
@@ -1,7 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`Footer renders correctly 1`] = `
-<Component
+<_temp2
```
ibeckermayer, line 4: Any idea why these changed?

colbymorrison: No I’m not sure.

colbymorrison:
@leon-i 

The Babel and console issues come from you using Windows. See line 17 of `frontend/webpack.common.js`, we're saying "don't transpile anything in node_modules except for the @fvilers package". However, this is using UNIX path syntax. So when run on Windows, the regex is interpreted as "transpile everything in node_modules", hence the Babel issues. So, you need to change Line 17 to `exclude: /node_modules(?!(\/|\\)@fvilers)/`. I needed to do this when testing on IE, I didn't realize one of us was using Windows to develop. Relevant webpack issue: https://github.com/webpack/webpack/issues/2031.

leon-i:
Ah ok, makes sense - thanks for clarifying. It doesn’t seem to be impacting anything other than the console messages, so if I’m the only one it affects I don’t really mind.

PR #533, opened 2020-08-31
ManageMembers for Mobile

leon-i:
Currently, we have a placeholder for the ManageMembers/ManageTeams page on mobile - this PR/draft aims to address that.  
  
Added a new `MobileMemberInfo` component, which takes the same data that would've converted into the table on desktop, and instead makes some mobile friendly components to compact the info for each team member. Here's what it looks like:  
  
mobile_manage_members  
  
Current issues/concerns:  
* Can't get the `:after` on the RoleSelector to work. Pathing via url in the scss file throws up errors.  
* Amount of info shown before clicking/tapping - should we also show role?  
* Styling - any suggestions on tidying it up/making it look nicer are appreciated.

### DIFF
--- a/frontend/client/src/components/MobileMemberInfo.js
+++ b/frontend/client/src/components/MobileMemberInfo.js
@@ -1,0 +1,45 @@
+import React from 'react'
+import RoleSelector from '../components/RoleSelector'
+import arrowRight from '../../assets/arrow-right.svg'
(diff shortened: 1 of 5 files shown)


ibeckermayer (commented):
Ignore, I'm just clearing out old review requests in my GitHub account.

PR #536, opened 2020-09-11
Timer Fix

colbymorrison:
[Asana Task](https://app.asana.com/0/1188301658055194/1192892309652977)

Updates the countdown timer to be its own component. Previously, the timer was created by a series of recursive function calls that updated the `timeLeft` state variable. However, when the reset button was pushed, the recursion was not stopped. So, when a new code was generated on the same page, there were two recursive chains running, each updating `timeLeft` on top of each other. This is shown below with the interval between updates changed from one minute to one second.

out

Now that the countdown timer is its own component, on reset the component is unmounted so the timer disappears. Also lets the time updates be triggered by the `useEffect` hook, which avoids the recursion. 

Now, with one second updates we get:
out2

### DIFF
--- a/frontend/client/src/components/Countdown.js
+++ b/frontend/client/src/components/Countdown.js
@@ -8,7 +8,7 @@ const Countdown = (props) => {
     timeLeft > 0 &&
       setTimeout(() => {
         setTimeLeft(timeLeft - 1)
-      }, 600)
+      }, 60000)
   }, [timeLeft])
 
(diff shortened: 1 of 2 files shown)


jcolla-holla (approved):

jcolla-holla:
Just tested locally including hitting the time when the code expires and the message changes.  Looks good, merging to staging for further testing.

